Issue #1, opened 2016-09-20
Add test for UTF-8 Parsing


jwilm:
The `utf8parse` package could use a test validating the implementation. My thought about how to do this is to find some [complicated UTF-8 file] and compare results of parsing it with this package's parser and the standard library's `std::str::from_utf8()`.

## Approach

Implement a `utf8parse::Receiver` for something wrapping a `String`. Read the UTF-8 test file into a buffer. for each byte in the buffer, `advance()` the `utf8parse::Parser`. After running all bytes through the parser, let the string be the `actual` value. Now, call `String::from_utf8()` on the test file buffer; let that be the `expected` value. Assert that those values are equal.

## Links

* [docs for utf8parse](https://docs.rs/utf8parse/0.1.0/utf8parse/index.html)
* [complicated UTF-8 file]

[complicated UTF-8 file]: https://raw.githubusercontent.com/tmux/tmux/master/tools/UTF-8-demo.txt

AravSinghal:
I love how you're adding easy issues for new contributors. Might be worth working on this project once I learn Rust. I need to create a very basic virtual terminal widget for a project anyway, and this project is the only recently updated project related to virtual terminals in Rust. Do you plan on making this a full blown virtual terminal like the inactive [rust-terminal](https://github.com/skade/rust-terminal)?

jwilm:
@AravSinghal this project is just intended to be a state machine for the ansi terminal protocol, nothing more. I have plans for higher level libraries as well which will use this project internally.

lizbaillie:
I would like to take this! Thanks for bringing it to my attention.

jwilm:
@lizbaillie all yours :)

lizbaillie:
@jwilm created a WIP PR - did not get very far this morning, but thought it might be helpful for you to see my progress so far, in case there was something I was doing obviously wrong from the get-go 😛

jwilm:
Resolved in #2

PR #2, opened 2016-10-05
adds test for UTF-8 parsing

lizbaillie:
- adds test as outlined in #1


jwilm:
This is exactly the right direction! Re: right place for the test; there's actually two crates here. There's the `vte` crate and the `utf8parse` crate. This test only involves the latter crate, so maybe it could all go under that tree?

lizbaillie:
No problem, can do! Thanks for the feedback, I will keep you posted on my progress.

jwilm:
I'm happy to give as much feedback as you like! Really, I'm just super jazzed that you decided to tackle this issue. Thanks :smile:

lizbaillie:
Started trying my hand at further implementation, but not really sure what I'm doing 😛 

Feel free to take a look at the most recent commit - I left comments about what I didn't understand. I can elaborate on any details if needed!

utf8parse/src/lib.rs
```diff
@@ -98,13 +98,30 @@ mod tests {
     use std::io::prelude::*;
     use std::fs::File;
 
+    // i have no idea what I'm doing here
+    impl Receiver for String {
```
jwilm: What you've got here is nearly correct. The `Receiver` type is called by the `utf8parse::Parser` whenever it completes a codepoint or an invalid sequence. There's an API [`String::push(&mut self, char)`](https://doc.rust-lang.org/std/string/struct.String.html?search=#method.push) which should allow you to build a string using these callbacks.

utf8parse/src/lib.rs
```diff
@@ -98,13 +98,30 @@ mod tests {
...
     fn utf8parse_test() {
         let mut buffer = String::new();
-        let file = File::open("UTF-8-demo.txt"); // complicated UTF-8 file
-        let file_to_string = file.read_to_string(&mut buffer);
-        let actual; // utf8parse implementation
-        let expected; // standard library implementation
+        let mut file = try!(File::open("UTF-8-demo.txt")); // complicated UTF-8 file
+        let file_to_string = try!(file.read_to_string(&mut buffer));
+        let parser = Parser::new();
+        let receiver; // ???
```
jwilm: You've got an `impl utf8parse::Receiver for String`, so the receiver would just be a `String::new()`.

utf8parse/src/lib.rs
```diff
@@ -98,13 +98,30 @@ mod tests {
...
+        let actual; // should this just be the buffer?
```
jwilm: This should be the `String` you build using the `utf8parse::Parser`.

lizbaillie:
Added a new commit with some failed efforts. Commented with the errors particular to the lines. I am a little tired so my brain is not working very well at the moment, but I figured I'd show you where I was at! Maybe it will make more sense to me in the morning 😛 

I'll squash all these when I'm done, just FYI 😄

utf8parse/src/lib.rs
```diff
@@ -91,3 +91,41 @@ impl Parser {
...
+        let mut file_to_string = file.read_to_string(&mut buffer).unwrap();
+        let parser = Parser::new();
+
+        // expected struct `std::vec::Vec`, found usize
+        let expected = String::from_utf8(file_to_string).unwrap();
+
+        // no method named `iter` found for type `usize` in the current scope
+        let actual = for byte in file_to_string.iter() {
+            // expected &mut _, found struct `std::string::String`
+            parser.advance(buffer, byte);
```
jwilm: There's not going to be a useful return value for this loop. The `actual` string you're producing _is_ the receiver that gets passed to `advance()`.

lizbaillie:
Thanks for the feedback! Sorry that I haven't been able to get back to this yet this week. Should be able to get back to it in a few days.

jwilm:
No worries! Take you time :)

lizbaillie:
Phew! Thanks for your patience. Okay, I think I'm a little further along, just having trouble figuring out exactly how to implement the `utf8parse` version. Any hints you might be able to offer on the most recent commit?

jwilm:
Yeah! Take a peak at the "Parser::advance" method signature in the docs. Specifically, check how it wants the receiver to be passed. On mobile right now so can't get you a link easily, but I think there's one up this thread somewhere.

lizbaillie:
Sounds good! I'll let you know if I have any further questions. Thanks!

jwilm:
I guess there wasn't a link in this thread yet. [`Parser::advance()`](https://docs.rs/utf8parse/0.1.0/utf8parse/struct.Parser.html#method.advance)

lizbaillie:
I think I _might_ have done it?? Let me know what you think! If it is in fact done, I can squash these commits.

jwilm:
Awesome! Let me grab your branch and give it a spin. Sadly I don't have travis set up yet because there was no tests up until now :flushed:

utf8parse/src/lib.rs
```diff
@@ -91,3 +91,43 @@ impl Parser {
...
+
+    #[test]
+    fn utf8parse_test() {
+        let mut buffer = String::new();
+        let mut file = File::open("src/UTF-8-demo.txt").unwrap();
+        let mut parser = Parser::new();
+
+        // assigning to _ so it won't matter that I'm not using the returned Result
+        // just want to read the file to a buffer
+        let _ = file.read_to_string(&mut buffer);
```
jwilm: We should probably call `unwrap()` or `expect()` on this result in case there actually is a problem; it would otherwise be more difficult to figure out where the test went wrong.

utf8parse/src/lib.rs
```diff
@@ -91,3 +91,43 @@ impl Parser {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::io::prelude::*;
```
jwilm: Can this import be changed to import only what is needed? I have an allergy to glob imports :wink:

lizbaillie:
Okay, all done! Let me know if there's anything else I need to do. 😄

jwilm:
This is awesome! Thank you thank you thank you :grin:

PR #11, opened 2018-01-11
Fix Tests

nixpulvis:
Links against STD for tests to allow use of `Vec`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,4 @@ rust:
   - nightly
 
 script:
-  - cd utf8parse && cargo test
+  - cargo test && cd utf8parse && cargo test
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,5 @@ rust:
   - nightly
 
 script:
-  - cargo test && cd utf8parse && cargo test
+  - cargo test
+  - cd utf8parse && cargo test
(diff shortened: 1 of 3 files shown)


nixpulvis:
Hmmm, interesting. I didn't look at the `.travis.yml`... out of curiousity why is it only testing the `utf8parse` crate?

Also I can look into fixing the tests in that crate as well, I'm assuming it's going to be mostly the same thing.

jwilm:
Thanks for working on this!

> out of curiousity why is it only testing the utf8parse crate?

Most likely a series of oversights

jwilm:
I think the reason only utf8parse is tested is that we didn't have any vte tests when travis was added. Should be easy enough to add to the .travis.yml, though.

.travis.yml
```diff
@@ -7,4 +7,4 @@ rust:
   - nightly
 
 script:
-  - cd utf8parse && cargo test
+  - cargo test && cd utf8parse && cargo test
```
jwilm: this should probably be split across two entries:

```yaml
script:
  - tests1
  - tests2
```

nixpulvis: I didn't even know you could do that! Done.

nixpulvis:
Everything passes now on travis... but I can't view the logs, it's giving me an error "There was an error while trying to fetch the log." so I'm not 100% sure that it's testing the vte crate itself now too.

jwilm:
I'm able to see the logs now.

This all looks great; thanks for sending the PR!

PR #12, opened 2018-01-12
Cargo Workspace

nixpulvis:
Not really a big deal, just a tiny little improvement. Since `vte` depends on `utf8parse` inside the same directory tree it's clearly a workspace, and this enables that. Other than `cargo test --all` I'm not really sure of all the advantages of using a workspace here, but still 😜.


jwilm:
thanks!

PR #16, opened 2018-08-11
Fix off-by-one error when parsing CSI escapes

chrisduerr:
The CSI escape sequences are limited to 16 parameters using the
`MAX_PARAMS` constant.

To make sure no more params are parsed, the CSI parser stops when the
parameter index reaches `MAX_PARAMS`, however this means that it stops
at index 16, which means that the number of params would be 17 since
arrays start at 0.

To fix this off-by-one error the CSI parser now stops when the index
reaches `MAX_PARAMS - 1`, which means it will stop when the index
reaches the last element in an array with `MAX_PARAMS` number of
elements.

This fixes jwilm/alacritty#1505.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -332,7 +332,7 @@ impl Parser {
                     // Completed a param
                     let idx = self.num_params;
 
-                    if idx == MAX_PARAMS {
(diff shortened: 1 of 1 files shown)


jwilm:
Published 0.3.3

PR #20, opened 2019-03-17
Use dynamically sized buffer for OSC

wez:
Whilst playing around with vte and processing iTerm2's image protocol,
I ran into the relatively small statically sized OSC buffer.

This diff switches the buffer from a static array to use a `Vec<u8>`.  I
did briefly experiment with just making this buffer much larger (as
https://github.com/jwilm/vte/pull/15 does), but even at 1MB (which is
too small for images) that size can lead to a stack overflow during
initialization.

In order to preserve building in a `no_std` environment this is done
conditionally.  I've followed the pattern used in other crates with
similar requirements:

* Introduced a `std` feature to signal that `std` is ok
* The `std` feature is on by default
* `no_std` crates that consume this one will need to set `vte` to
  `default-features = false` in their deps
* I've bumped the version in order to avoid breaking `no_std`
  consumers

Test Plan:

```
$ cargo test --no-default-features
$ cargo test
```

Refs: https://github.com/jwilm/alacritty/issues/1002

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,7 +94,7 @@ mod oscbuf {
         pub fn new(size_limit: usize) -> Self {
             Self {
                 buf: Vec::with_capacity(MAX_OSC_RAW.min(size_limit)),
-                    size_limit,
+                size_limit,
             }
         }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,11 +31,11 @@
 //! [`Perform`]: trait.Perform.html
 //! [Paul Williams' ANSI parser state machine]: https://vt100.net/emu/dec_ansi_parser
 
-#![cfg_attr(not(feature="std"), no_std)]
+#![cfg_attr(feature="nostd", no_std)]
 
 extern crate utf8parse as utf8;
 
-#[cfg(feature="std")]
+#[cfg(not(feature="nostd"))]
 extern crate core;
 
 use core::mem;
@@ -65,13 +65,13 @@ impl State {
     }
 }
 
-#[cfg(feature="std")]
+#[cfg(not(feature="nostd"))]
 mod oscbuf {
     use super::MAX_OSC_RAW;
     pub const DEFAULT_SIZE_LIMIT: usize = u32::max_value() as usize;
 
     /// When std is available, we can simply use a Vec for the
-    /// OSC storage.  We'll pre-size it to something reasonable.
+    /// OSC storage. We'll pre-size it to something reasonable.
     pub struct OscBuffer {
         buf: Vec<u8>,
         size_limit: usize,
@@ -99,33 +99,32 @@ mod oscbuf {
         }
 
         pub fn clear(&mut self) {
-            // If we're a bit heavy, shrink our usage.
+            // If we're a bit heavy, make an attempt to shrink our usage.
             // Ideally we'd call Vec::shrink_to() but that is
             // nightly only, so we fake it by resizing and then
             // asking for a shrink to fit the desired capacity.
             if self.buf.capacity() > 2 * MAX_OSC_RAW {
-                self.buf.resize(MAX_OSC_RAW, 0u8);
+                self.buf.truncate(MAX_OSC_RAW);
                 self.buf.shrink_to_fit();
             }
             self.buf.clear();
         }
 
         /// Returns true if we are able to store an additional
-        /// byte of data.  Since we have an allocator available,
-        /// we always return true.
-        pub fn have_capacity(&self) -> bool {
+        /// byte of data.
+        pub fn has_capacity(&self) -> bool {
             self.buf.len() < self.size_limit
         }
     }
 }
 
-#[cfg(not(feature="std"))]
+#[cfg(feature="nostd")]
 mod oscbuf {
     use super::MAX_OSC_RAW;
     pub const DEFAULT_SIZE_LIMIT: usize = MAX_OSC_RAW;
     /// The no_std version of OscBuffer uses a statically allocated
     /// array and thus has a limit on the size of the data that it
-    /// can hold.  If an OSC exceeds this length, it will be silently
+    /// can hold. If an OSC exceeds this length, it will be silently
     /// truncated.
     /// OscBuffer presents a subset of the Vec API.
     pub struct OscBuffer {
@@ -136,7 +135,7 @@ mod oscbuf {
 
     impl OscBuffer {
         pub fn new(size_limit: usize) -> Self {
-            debug_assert!(size_limit <= MAX_OSC_RAW);
+            assert!(size_limit <= MAX_OSC_RAW);
             Self {
                 buf: [0; MAX_OSC_RAW],
                 idx: 0,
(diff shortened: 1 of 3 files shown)


jwilm:
Thanks for the great PR; I love the approach here! One of the things that was nice about the fixed-size buffer was that a misbehaving terminal application couldn't cause the vte process to allocate an unlimited amount of memory.

How do you think about having an "unlimited" sized buffer as in this PR? On one hand, it's nice being able to opt into a limit to protect resources and other processes on the machine (ugh, oom-killer). On the other hand, TEs are usually running trusted applications, and the main avenue for excessive consumption would be through misuse of an application.

wez:
I suspect that the most likely cause of a problem with this is going to be a programmer error rather than something malicious; when I think of malicious terminal exploits I think largely of nefarious traps in web pages when using copy and paste, rather than running an untrusted application via the terminal--in the untrusted application scenario, the primary problem is the running of that application.

Having a limit to protect against something stupid doesn't sound unreasonable, but I can also see it potentially getting in the way.  For example, the iTerm image protocol is also able to be used to transfer files of arbitrary size (https://www.iterm2.com/documentation-images.html) so setting the limit too small could break this.

Here's what I suggest: introduce a configurable limit that can be set when constructing the parser.  The default could be say `u32::max_value()` to set a ballpark upper bound and allow the parser to eventually make progress.  That limit is big enough for a lot of things and not too big that it would on its own exhaust all the memory on the machine, and hopefully keep things under the radar from eg: oom killer. 

We could also potentially augment the `clear` call to resize the vec back down to 1k if it is above some threshold to keep the heap size better constrained.

jwilm:
A configurable max sounds good to me. I also like the idea of freeing the memory after use. Thanks for the feedback!

wez:
bump

jwilm:
Hey @wez, thanks for the bump, and sorry for the delay here. I'm in the process of moving this week won't be able to get back to this for _at least_ a week or two.

@chrisduerr may be able to take a peak at this sooner than that (no promises).

Thanks for your patience.

chrisduerr:
I can add some comments if a review is desired, however I don't have access to this repo so I can't merge it if it does not have any issues.

wez:
friendly ping :)

chrisduerr:
I'm currently busy with https://github.com/jwilm/alacritty/pull/2438 which will probably take another week.

I have this tracked, but it will take some time until I can take a look at it.

src/lib.rs
```diff
@@ -61,9 +65,116 @@ impl State {
...
+
+#[cfg(not(feature="std"))]
+mod oscbuf {
+    use super::MAX_OSC_RAW;
+    pub const DEFAULT_SIZE_LIMIT: usize = MAX_OSC_RAW;
+    /// The no_std version of OscBuffer uses a statically allocated
+    /// array and thus has a limit on the size of the data that it
+    /// can hold.  If an OSC exceeds this length, it will be silently
+    /// truncated.
+    /// OscBuffer presents a subset of the Vec API.
```
nixpulvis: Almost seems like [`FixedVec`](https://docs.rs/fixedvec/0.2.3/fixedvec) could be nice, but it looks like it might not be better, or at least not yet.

src/lib.rs
```diff
@@ -61,9 +65,116 @@ impl State {
...
+        pub fn push(&mut self, byte: u8) {
+            self.buf[self.idx] = byte;
+            self.idx += 1;
+        }
+
+        /// Returns true if we are able to store an additional
+        /// byte of data.  Since we have no allocator available,
+        /// we only return true while we are within the bounds
+        /// of the static storage.
+        pub fn have_capacity(&self) -> bool {
```
chrisduerr: I think `has_capacity` is probably the better name here.

src/lib.rs
```diff
@@ -61,9 +65,116 @@ impl State {
...
+    /// OscBuffer presents a subset of the Vec API.
+    pub struct OscBuffer {
+        buf: [u8; MAX_OSC_RAW],
+        idx: usize,
+        size_limit: usize,
+    }
+
+    impl OscBuffer {
+        pub fn new(size_limit: usize) -> Self {
+            debug_assert!(size_limit <= MAX_OSC_RAW);
```
chrisduerr: If we have a limitation like this, it should be a runtime Error/hard panic in all modes or safety should be ensured using the typesystem (like using a u16 instead).

However, I'm not sure if we need to do that here. You said that for a stack implementation you'll run out of stack space anyways for bigger stuff, so why not just stick to the old length of `1024` and hide this complexity from the consumer who will likely have no idea which effects this is going to have?

src/lib.rs
```diff
@@ -211,7 +324,7 @@ impl Parser {
 
         for i in 0..self.osc_num_params {
             let indices = self.osc_params[i];
-            slices[i] = &self.osc_raw[indices.0..indices.1];
+            slices[i] = &self.osc_raw.as_slice()[indices.0..indices.1];
```
chrisduerr: This `as_slice` is unnecessary, Vecs will be automatically sliced to slices.

src/lib.rs
```diff
@@ -61,9 +65,116 @@ impl State {
     }
 }
 
+#[cfg(feature="std")]
+mod oscbuf {
+    use super::MAX_OSC_RAW;
+    pub const DEFAULT_SIZE_LIMIT: usize = u32::max_value() as usize;
+
+    /// When std is available, we can simply use a Vec for the
+    /// OSC storage.  We'll pre-size it to something reasonable.
```
chrisduerr: It seems like you've put double-spaces behind all full stop in comments. These should all be single spaces following the guideline set by existing comments.

This is true for all comments, I've annotated only this one for convenience.

src/lib.rs
```diff
@@ -61,9 +65,116 @@ impl State {
...
+    pub struct OscBuffer {
+        buf: Vec<u8>,
+        size_limit: usize,
```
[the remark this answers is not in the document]
wez: https://github.com/jwilm/vte/pull/20#issuecomment-473946665

src/lib.rs
wez: I'll make this an unconditional `assert!`.  Anything else propagates up through the API in a way that I think will be ugly and take more wrangling to resolve than I think is needed.

The size limit needs to be passed down and respected in order for the unit tests to pass in both std and nostd feature modes.

src/lib.rs
wez: `as_slice` is necessary because `nostd` doesn't have `Vec` and so `OscBuffer` doesn't automatically slice in that way; removing `as_slice` causes the tests to fail in the `no_std` case.

src/lib.rs
chrisduerr: > The size limit needs to be passed down and respected in order for the unit tests to pass in both std and nostd feature modes.

I don't think it makes sense to have a worse API just to have tests complete in both modes. I'd say that runtime errors are usually extremely annoying for consumers, so I think it's probably best to stay away from them whenever possible.

wez: I understand your perspective, but I'm not clear what it is that you want to be done about it.  Can you give specific advice on how to resolve this?

wez: Please also keep in mind that this whole size limit thing is borne out of code review here; it was not in my initial implementation and it is pretty frustrating that you're not in harmony with each other on this point.

src/lib.rs
```diff
@@ -61,9 +65,116 @@ impl State {
     }
 }
 
+#[cfg(feature="std")]
+mod oscbuf {
+    use super::MAX_OSC_RAW;
+    pub const DEFAULT_SIZE_LIMIT: usize = u32::max_value() as usize;
```
chrisduerr: If we're concerned about a DOS attack by applications running inside the terminal, it seems like `u32::max_value()` is a bad value to choose here.

PR #21, opened 2019-04-24
Use correct md code fence

felixrabe:
Preview:

```sh
cargo build --release --example parselog
vim | target/release/examples/parselog
```


jwilm:
thanks!

PR #25, opened 2019-08-06
Add an implicit zero after trailing semicolon

### LINKED ISSUE #23: Interpret a trailing semicolon as an implicit '0' parameter
From what I can find, escape sequence parameters are "optional numeric values separated by semicolons".

As the values are separated by semicolons, it makes sense that if a sequence ends with a semicolon, the terminal should expect another parameter. Because the parameter is an optional number, it should default to 0.

For example, the code `\e[4;m` should have the same effect as `\e[4;0m`, that is, none at all.

I tested this specific sequence in kitty, termite, xterm, st and alacritty, and only alacritty draws the following text underlined.

It should be a simple fix and I will include it in my PR for #22

VojtechStep:
fixes #23

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,7 +89,6 @@ pub struct Parser {
     intermediate_idx: usize,
     params: [i64; MAX_PARAMS],
     param: i64,
-    collecting_param: bool,
     num_params: usize,
     osc_raw: [u8; MAX_OSC_RAW],
     osc_params: [(usize, usize); MAX_PARAMS],
@@ -108,7 +107,6 @@ impl Parser {
             intermediate_idx: 0,
             params: [0i64; MAX_PARAMS],
             param: 0,
-            collecting_param: false,
             num_params: 0,
             osc_raw: [0; MAX_OSC_RAW],
             osc_params: [(0, 0); MAX_PARAMS],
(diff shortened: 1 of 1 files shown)


VojtechStep:
Seems alright and checks pass, let me change it real quick.

src/lib.rs
```diff
@@ -294,11 +292,10 @@ impl Parser {
             },
             Action::Unhook => performer.unhook(),
             Action::CsiDispatch => {
-                if self.collecting_param {
-                    let idx = self.num_params;
-                    self.params[idx] = self.param;
-                    self.num_params += 1;
-                }
+
```
chrisduerr: There's an excess newline here.

src/lib.rs
```diff
@@ -557,6 +551,19 @@ mod tests {
 
     }
 
+    #[test]
+    fn parse_trailing_semi() {
```
chrisduerr: I know that you've probably used the other tests to come up with this name, but I think it's a bit unclear.

I feel like `csi_parse_trailing_semicolon` might be a bit clearer here.

VojtechStep: Sure, I was just trying to pick up on the convention :)

chrisduerr: > Sure, I was just trying to pick up on the convention :)

Yeah, which I think was a good idea. But I feel like the conventions used here aren't optimal here.

jwilm:
Sweet PR! Gotta love that some branching and state were removed. :smile:

PR #26, opened 2019-08-07
Discard CSI sequences with too many parameters

### LINKED ISSUE #24: Decide on the correct way of handling excess control sequence parameters
In the current state, a control sequence that is supplied more than the maximum number of parameters weirdly smashes the trailing ones into a big number, behaving as if the parameter separators didn't exist. For example, the control sequence `\e[38;2;0;255;0;48;2;255;0;0;1;2;3;4;5;6;0mHello` gets parsed as `[38, 2, 0, 255, 0, 48, 2, 255, 0, 0, 1, 2, 3, 4, 5, 60]` - notice the 60 at the end.

AFAIK, the standard doesn't define any maximum number of parameters. Since I couldn't find any canonical way handle the excess of parameters, this is the behavior of other terminal emulators (data collected mostly by trying it on my machine and browsing code):

- Trash the entire control sequence - st, kitty, GNOME VTE (kitty seems to then print the sequence as normal printable characters)
- Trash the excess parameters - apparently DEC VT100+ compatible emulators do this
- After reaching the maximum capacity, instead of appending, overwrite the last parameter with new ones - Konsole (not a great solution, however their limit is 4096 parameters, so you won't run into this all that often)

(Couldn't figure out what xterm does, seems to me like it just craps its pants and doesn't give a damn)

I'm inclined to implement the first option, but I don't have a strong opinion on it. The current way just doesn't seem "correct" in any shape or form and the VTE crate is the only one where I've seen this behavior so far.

VojtechStep:
CSI sequences that don't fit into the internal buffer don't get reported as valid sequences to the Performer.
Instead, such cases get reported as a warning, which is handled based on the calling binary's usage of different available frontends for the `log` crate.

Note that we cannot print the entire invalid sequence, because the rest of the parameters aren't in the buffer, and if they were there would be no reason for this PR, so only the first 16 parameters are reported.

Fixes #24

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -375,8 +375,8 @@ pub trait Perform {
 
     /// A final character has arrived for a CSI sequence
     ///
-    /// The `ignore` flag indicates that more than two intermediates arrived and
-    /// subsequent characters were ignored.
+    /// The `ignore` flag indicates that either more than two intermediates arrived
+    /// or the number of parameters exceeded MAX_PARAMS, and subsequent characters were ignored.
     fn csi_dispatch(&mut self, params: &[i64], intermediates: &[u8], ignore: bool, _: char);
 
     /// The final character of an escape sequence has arrived.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -272,8 +272,12 @@ impl Parser {
             },
             Action::Unhook => performer.unhook(),
             Action::CsiDispatch => {
-                self.params[self.num_params] = self.param;
-                self.num_params += 1;
+                if self.num_params == MAX_PARAMS {
+                    self.ignoring = true;
+                } else {
+                    self.params[self.num_params] = self.param;
+                    self.num_params += 1;
+                }
 
                 performer.csi_dispatch(
                     self.params(),
@@ -298,14 +302,15 @@ impl Parser {
                 }
             },
             Action::Param => {
+                // Completed a param
+                let idx = self.num_params;
+
+                if idx == MAX_PARAMS {
+                    self.ignoring = true;
+                    return;
+                }
+
                 if byte == b';' {
-                    // Completed a param
-                    let idx = self.num_params;
-
-                    if idx == MAX_PARAMS - 1 {
-                        return;
-                    }
-
                     self.params[idx] = self.param;
                     self.param = 0;
                     self.num_params += 1;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -613,6 +613,7 @@ mod tests {
         assert!(dispatcher.dispatched_csi);
         assert_eq!(dispatcher.params.len(), 1);
         assert_eq!(dispatcher.params[0].len(), MAX_PARAMS);
+        assert!(!dispatcher.ignore);
     }
 
     #[test]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -376,7 +376,8 @@ pub trait Perform {
     /// A final character has arrived for a CSI sequence
     ///
     /// The `ignore` flag indicates that either more than two intermediates arrived
-    /// or the number of parameters exceeded MAX_PARAMS, and subsequent characters were ignored.
+    /// or the number of parameters exceeded the maximum supported length,
+    /// and subsequent characters were ignored.
     fn csi_dispatch(&mut self, params: &[i64], intermediates: &[u8], ignore: bool, _: char);
 
     /// The final character of an escape sequence has arrived.
@@ -395,6 +396,7 @@ mod tests {
     use super::*;
 
     use core::i64;
+    use std::string::String;
     use std::vec::Vec;
 
     static OSC_BYTES: &[u8] = &[
@@ -567,8 +569,6 @@ mod tests {
 
     #[test]
     fn parse_osc_max_params() {
-        use crate::MAX_PARAMS;
-
         static INPUT: &[u8] = b"\x1b];;;;;;;;;;;;;;;;;\x1b";
 
         // Create dispatcher and check state
@@ -591,12 +591,10 @@ mod tests {
 
     #[test]
     fn parse_csi_max_params() {
-        use crate::MAX_PARAMS;
-
         // This will build a list of repeating '1;'s
         // The length is MAX_PARAMS - 1 because the last semicolon is interpreted
         // as an implicit zero, making the total number of parameters MAX_PARAMS
-        let params = std::iter::repeat("1;").take(MAX_PARAMS - 1).collect::<std::string::String>();
+        let params = std::iter::repeat("1;").take(MAX_PARAMS - 1).collect::<String>();
         let input = format!("\x1b[{}p", &params[..]).into_bytes();
 
         // Create dispatcher and check state
@@ -618,12 +616,10 @@ mod tests {
 
     #[test]
     fn parse_csi_params_ignore_long_params() {
-        use crate::MAX_PARAMS;
-
         // This will build a list of repeating '1;'s
         // The length is MAX_PARAMS because the last semicolon is interpreted
         // as an implicit zero, making the total number of parameters MAX_PARAMS + 1
-        let params = std::iter::repeat("1;").take(MAX_PARAMS).collect::<std::string::String>();
+        let params = std::iter::repeat("1;").take(MAX_PARAMS).collect::<String>();
         let input = format!("\x1b[{}p", &params[..]).into_bytes();
 
         // Create dispatcher and check state
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -450,6 +450,7 @@ mod tests {
     #[derive(Default)]
     struct CsiDispatcher {
         dispatched_csi: bool,
+        ignore: bool,
         params: Vec<Vec<i64>>,
     }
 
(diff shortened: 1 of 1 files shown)


chrisduerr:
I've wanted to play around with an alternative implementation for a bit, but once I was done I had already made some pretty significant changes. So I thought the best way of sharing would be to just push to this PR. Please note that I'm not trying to make a definite statement on what this PR should be here, this is simply my proposal for improving on your work. With the powers of git we could still just revert the changes if we would wish to do so.

Now to explain the changes I have made. The most uncontroversial stuff I've changed is probably that I've fixed all clippy issues and I refactored tests a bit and added a test specifically to check that `MAX_PARAMS` itself was still supported, not only that `MAX_PARAMS + 1` is not.

I've also kinda switched up your logic a bit. Instead of checking for `MAX_PARAMS` and then setting a flag that we've reached the limit, I now just check `num_params` to see if we should throw away the escape. We can do this since we will always push a new escape sequence. If the escape ended with `;`, we will push a new one with `0` before dispatching, otherwise we will dispatch the queued escape. This allows us to just fill up the params array all the way and check if we've reached our limit before dispatching, getting rid of the unnecessary variable.

Now for the thing I'm most conflicted about. I've removed the `log` dependency. This has various reasons.

First of all, it's a depedency to something that has worked completely without depedencies for now (utf8parse is part of the repo). For me personally this is not a good enough reason alone to block it, however it raises the bar a bit. Going from no external dependencies to one will mean we have to continually update vte to make sure that we do not cause duplicate dependencies downstream in Alacritty.

Another reason is that currently no child logs are printed in Alacritty at all. This is because there's just so much stuff logged, that it would trash Alacritty's log output. So if we just make this change, it would require explicit downstream patching to actually get any output of these errors.

The most significant argument though is that I question how useful this message would be. We don't actually want it as a warning, since that would pop open the message bar and we do not do that for unsupported escapes. What we would want to do is just log it. However these messages are only logged when logging is explicitly enabled, which even for debugging is rarely the case.

So the only scenario where this would be useful is if there is an application using more than 16 parameters (already unlikely), which causes a bug that directly impacts the user enough to notice and report it, leading to us guessing that this might be the issue (at which point we could probably just check manually), asking them to provide a debug log and then finding this message to identify and confirm the issue.

I wouldn't bet on that never happening, however I think the chances of this ever helping us are extremely, extremely slim. So with that considered and taking all the other factors into account, I think our best bet is to just ignore the escape and keep going, assuming that whenever this happens, it's either malicious or an accident.

However I can see the other side of it as well, so I'd be curious what @nixpulvis thinks too.

VojtechStep:
I wasn't sure about the logging either, but because all the other emulators I worked with logged it, I thought we might as well.

However I agree that it adds maintenance cost and it would be nice to stay without dependencies.

Add it behind a feature flag maybe?

chrisduerr:
> Add it behind a feature flag maybe?

Since Alacritty is pretty much the only consumer of this crate for whom this is relevant, I don't think it makes sense to add features unless they're used by Alacritty. At least not unless they're explicitly requested.

VojtechStep:
Among the crates dependent on vte is a repository called `ransid`, which AFAIK is a terminal emulator for RedoxOS and has over 10,000 downloads on crates.io. It also uses the log crate and logs invalid control sequences in debug, so they might be interested in such a feature.

chrisduerr:
Orbterm is a decent point, however I still think if we make it a feature it should be requested if desired, rather than just providing it with the expectation that someone might potentially need it.

jwilm:
One alternative to logging here directly would be to add another `Perform` method which is passed the garbage data instead of the csi dispatch method. If the implementor doesn't want to log or anything (ie blank impl), then it just gets compiled out. We could log if desired in Alacritty.

chrisduerr:
Actually, now that you're mentioning it, couldn't we just set `ignoring` to `true`? It seems like the purpose of `ignoring` is kinda similar this.

It might also make it possible to leave it up to the application if it wants to discard the escape sequence or if it wants to parse whatever it got. So that could actually be a big advantage.

nixpulvis:
The `log` crate is somewhat standard, but this would allow users of this library to always choose how they want to handle this case, be it a `log`, `println!`, or anything else. Given @chrisduerr's point about going from effectively 0 dependencies to 1 is a significant change, even if the `log` crate is widely used.

As for `ignoring`, it seems that's currently only used for `intermediates`. So this might work, but it would be changing the meaning to include ignoring params as well. What issues might this cause in client code? Unless I'm missing something, Alacritty doesn't use this value for anything but calls to `unhandled!` currently.

chrisduerr:
> As for ignoring, it seems that's currently only used for intermediates. So this might work, but it would be changing the meaning to include ignoring params as well. What issues might this cause in client code? Unless I'm missing something, Alacritty doesn't use this value for anything but calls to unhandled! currently.

It would certainly be a breaking change. But Alacritty basically just ignores the escape sequence when `ignoring` is set.

This certainly would be a big change in what `ignoring` does, but I think usually you would want to handle these two very similarly. The only big issue would be that you currently can tell exactly why `ignoring` was set and if we would use it for multiple things, it could be any of those things causing the flag to be set.

I'm not quite certain where the `ignoring` flag came from. I'm not sure if this is a standard thing with a commonly known name or if it was just made for this purpose in the API of the VTE crate.

If this is just a VTE thing, @jwilm's approach might actually improve the API a bit and allow us to remove all `ignoring` flags from all the methods in the crate and instead have just one 'catch-all' error handler which can be notified with a specific cause about why an escape couldn't be parsed.

VojtechStep:
Looking at the description of the state machine this crate is based on ([link](https://vt100.net/emu/dec_ansi_parser#STCSIIGN)), the ignore state is meant to be used for exactly this purpose: to consume the rest of a sequence that should be considered invalid.

Obviously because the standard does not set any limit to the number of parameters, there is no mention of what to do with long sequences, but I feel like we've reached the consensus that it's an invalid sequence.

chrisduerr:
> Looking at the description of the state machine this crate is based on (link), the ignore state is meant to be used for exactly this purpose: to consume the rest of a sequence that should be considered invalid.

This is exactly what I was looking for. So as I've suspected the `ignore` flag is a common thing that probably shouldn't be abused too hard.

However it does sound a lot better than I had feared. Following this doc, it seems to hit the nail exactly on its head, we'd just have to fix our documentation.

The biggest problem then would be to clearly communicate this as a breaking change. Since people will likely be unaware even if we bump the version number accordingly.

VojtechStep:
@chrisduerr I rebased on master, implemented the functionality using the `ignore` flag, added relevant test cases and extended the documentation for `csi_dispatch`.
Please let me know if there is anything else for me to do.

src/lib.rs
```diff
@@ -370,8 +375,8 @@ pub trait Perform {
 
     /// A final character has arrived for a CSI sequence
     ///
-    /// The `ignore` flag indicates that more than two intermediates arrived and
-    /// subsequent characters were ignored.
+    /// The `ignore` flag indicates that either more than two intermediates arrived
+    /// or the number of parameters exceeded MAX_PARAMS, and subsequent characters were ignored.
```
chrisduerr: Since `MAX_PARAMS` is not a public constant, I'm not sure if we should use it in our documentation.

Maybe just a more unspecific `exceeded the maximum number of allowed parameters` is less confusing and hopefully doesn't send people on a documentation treasure hunt for this magical `MAX_PARAMS` constant.

src/lib.rs
```diff
@@ -586,22 +593,54 @@ mod tests {
     fn parse_csi_max_params() {
         use crate::MAX_PARAMS;
 
-        static INPUT: &[u8] = b"\x1b[1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;p";
+        // This will build a list of repeating '1;'s
+        // The length is MAX_PARAMS - 1 because the last semicolon is interpreted
+        // as an implicit zero, making the total number of parameters MAX_PARAMS
+        let params = std::iter::repeat("1;").take(MAX_PARAMS - 1).collect::<std::string::String>();
```
chrisduerr: This line got pretty long, if we extract std::iter and std::string::String as imports it should be a bit easier to read.

I'd assume the full `String` path is necessary because of the `#![no_std]` by default, in which case it might make sense to import `use std::prelude::v1::*;`, since that automatically imports everything we get in a normal `std` environment.

src/lib.rs
```diff
@@ -586,22 +593,54 @@ mod tests {
...
-        // Run parser using OSC_BYTES
+        // Run parser using INPUT
         let mut parser = Parser::new();
-        for byte in INPUT {
-            parser.advance(&mut dispatcher, *byte);
+        for byte in input {
+            parser.advance(&mut dispatcher, byte);
+        }
+
+        // Check that flag is set and thus csi_dispatch assertions ran.
```
chrisduerr:
```suggestion
        // Check that flag is set and thus csi_dispatch assertions ran
```

For consistency I'd remove the dot at the end here, since all the other comments don't seem to have one either.

src/lib.rs
```diff
@@ -586,22 +593,54 @@ mod tests {
...
+        assert!(dispatcher.dispatched_csi);
+        assert_eq!(dispatcher.params.len(), 1);
+        assert_eq!(dispatcher.params[0].len(), MAX_PARAMS);
+        assert!(!dispatcher.ignore);
+    }
+
+    #[test]
+    fn parse_csi_params_ignore_long_params() {
+        use crate::MAX_PARAMS;
```
chrisduerr: We have this in multiple tests now, but I don't see why we wouldn't add it at the top of the `test` module. Using just `use super::*;` would be fine too, since that's a pretty common pattern for test modules (at least in the repos of the Alacritty project).

VojtechStep:
I addressed most of your points, except for the dot in tests. I only copied what was in the existing test case above and IMO there should be another PR for unifying how the tests look, as I see it as out of scope for this one. I will probably look into it during the holiday.

PR #29, opened 2019-11-03
fix dcs handling, and add a test for it

doy:
fixes https://github.com/jwilm/vte/issues/28.

it turns out that there were a couple different issues with dcs parsing, and this should resolve all of them:

* the final character of the dcs introducer was not being passed to the `hook` method, making it impossible to determine which dcs sequence was being processed (which is necessary to correctly handle the `put` method)
* the last parameter was being incorrectly dropped

the test tests the dcs sequence `DECUDK` documented in the xterm documentation.

additionally, you can see the difference in the `parselog` output before this change:

```
$ perl -E'print "foo\eP0;1|17/ab\x9cbar"' | cargo run -q --example parselog
[print] 'f'
[print] 'o'
[print] 'o'
[hook] params=[0], intermediates=[], ignore=false
[put] 31
[put] 37
[put] 2f
[put] 61
[put] 62
[unhook]
[print] 'b'
[print] 'a'
[print] 'r'
```

compared to after:

```
$ perl -E'print "foo\eP0;1|17/ab\x9cbar"' | cargo run -q --example parselog
[print] 'f'
[print] 'o'
[print] 'o'
[hook] params=[0, 1], intermediates=[], ignore=false, char='|'
[put] 31
[put] 37
[put] 2f
[put] 61
[put] 62
[unhook]
[print] 'b'
[print] 'a'
[print] 'r'
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -186,13 +186,13 @@ impl Parser {
             state => {
                 // Exit action for previous state
                 let exit_action = self.state.exit_action();
-                maybe_action!(exit_action, 0);
+                maybe_action!(exit_action, byte);
 
                 // Transition action
                 maybe_action!(action, byte);
 
                 // Entry action for new state
-                maybe_action!(state.entry_action(), 0);
+                maybe_action!(state.entry_action(), byte);
 
                 // Assume the new state
                 self.state = state;
@@ -223,10 +223,14 @@ impl Parser {
             Action::Print => performer.print(byte as char),
             Action::Execute => performer.execute(byte),
             Action::Hook => {
+                self.params[self.num_params] = self.param;
+                self.num_params += 1;
+
                 performer.hook(
                     self.params(),
                     self.intermediates(),
                     self.ignoring,
+                    byte as char,
                 );
             },
             Action::Put => performer.put(byte),
@@ -378,7 +382,7 @@ pub trait Perform {
     ///
     /// The `ignore` flag indicates that more than two intermediates arrived and
     /// subsequent characters were ignored.
-    fn hook(&mut self, params: &[i64], intermediates: &[u8], ignore: bool);
+    fn hook(&mut self, params: &[i64], intermediates: &[u8], ignore: bool, char);
 
     /// Pass bytes as part of a device control string to the handle chosen in `hook`. C0 controls
     /// will also be passed to the handler.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -186,13 +186,13 @@ impl Parser {
...
                 let exit_action = self.state.exit_action();
-                maybe_action!(exit_action, 0);
+                maybe_action!(exit_action, byte);
 
                 // Transition action
                 maybe_action!(action, byte);
 
                 // Entry action for new state
-                maybe_action!(state.entry_action(), 0);
+                maybe_action!(state.entry_action(), byte);
```
chrisduerr: Just got a comment, not an issue for documentation purposes about this. I thought this would obviously cause a change in functionality, however there's a good reason why this change is not a problem:

There are only a limited amount of exit and entry actions, these include `Action::None`, `Action::Clear`, `Action::Hook`, `Action::OscStart`, `Action::Unhook` and `Action::OscEnd`.

Looking at the `perform_action` function now, we can see which action even consider the `byte` parameter. The parameter is used by `Action::Print`, `Action::Execute`, `Action::Put`, `Action::OscPut`, `Action::CsiDispatch`, `Action::EscDispatch`, `Action::Collect`, `Action::Param` and `Action::BeginUtf8`.

Comparing these two lists, it becomes clear that the parameter is currently unused entirely when `0` is passed. So passing `byte` as parameter and using it in `Action::Hook`, which previously did not use `byte`, will not create any conflicts.

src/lib.rs
```diff
@@ -465,6 +469,32 @@ mod tests {
...
+            self.s.push(byte);
+        }
+        fn unhook(&mut self) {
+            self.dispatched_dcs = true;
+        }
+        fn osc_dispatch(&mut self, _params: &[&[u8]]) { }
+        fn csi_dispatch(&mut self, _params: &[i64], _intermediates: &[u8], _ignore: bool, _c: char) {}
+        fn esc_dispatch(&mut self, _params: &[i64], _intermediates: &[u8], _ignore: bool, _byte: u8) {}
+    }
+
```
chrisduerr: The only actual issue I have found with this PR is this double newline right here. Other than that, things are looking good to me.

doy: fixed!

PR #33, opened 2019-11-22
Add dynamic buffer support

chrisduerr:


chrisduerr:
I ran performance benchmarks against Alacritty and was able to confirm that this does not impact performance.

A test with the clipboard escape also allowed me to confirm, that the downstream bug https://github.com/jwilm/alacritty/issues/1002 is indeed fixed by this.

PR #35, opened 2019-11-25
Remove table generation

chrisduerr:
This completely removes the `codegen` project, which relied on outdated
libraries to parse DSLs to build the utf8 and vte state tables, to make
the library easier to maintain.

The utf8 table could be completely removed in favor of a `match`
statement, which also lead to a performance improvement with the utf8
parser.

The vte table did not benefit from `match` statements at all and instead
had significantly worse performance with it. To replace the old
codegeneration for vte, the `generate_state_changes` crate has been
created instead, which uses the language's proc_macro feature to create
a `const fn` which will generate the table at compile time.


jwilm:
Proc macros weren't out yet at the time I built codegen, and it's good to see this getting modernized. Nice work :smile:

chrisduerr:
@jwilm I actually tried a lot of different ways to go without a proc macro too, similarly to the change you've made to improve the UTF-8 parser, but interestingly enough, for a big match like this there was no way to match the array performance. I even tried writing out all 2000+ statements by themselves in the match, since apparently ranges prevent the match from getting compiled to a jump table.

PR #44, opened 2020-01-28
Pass terminator to osc dispatcher

chrisduerr:
Even though the ST terminator is the only officially supported
terminator, some applications still rely on BEL to work properly. Both
have been supported historically, however there was no way for the
terminal to tell which terminator was used.

Since OSC escapes frequently offer the `?` parameter to query for the
current format, some applications expect the response terminator to
match the request terminator. To make it possible to support this, the
osc dispatcher is now informed when the BEL terminator was used.

Since the C1 ST terminator was not yet supported for OSC escapes,
support for it has also been added.

### DIFF
--- a/src/definitions.rs
+++ b/src/definitions.rs
@@ -50,7 +50,7 @@ pub enum Action {
 
 impl State {
     #[inline]
-    pub fn entry_action(&self) -> Action {
+    pub fn entry_action(self) -> Action {
         match self {
             State::CsiEntry | State::DcsEntry | State::Escape => Action::Clear,
             State::DcsPassthrough => Action::Hook,
@@ -60,7 +60,7 @@ impl State {
     }
 
     #[inline]
-    pub fn exit_action(&self) -> Action {
+    pub fn exit_action(self) -> Action {
         match self {
             State::DcsPassthrough => Action::Unhook,
             State::OscString => Action::OscEnd,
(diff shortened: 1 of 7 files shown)


src/definitions.rs
```diff
@@ -50,7 +50,7 @@ pub enum Action {
 
 impl State {
     #[inline]
-    pub fn entry_action(&self) -> Action {
+    pub fn entry_action(self) -> Action {
```
kchibisov: Why are we changing `&self` to `self`?

src/table.rs
```diff
@@ -170,6 +170,8 @@ generate_state_changes!(state_changes, {
         0x08..=0x17 => (Anywhere, Ignore),
         0x19        => (Anywhere, Ignore),
         0x1c..=0x1f => (Anywhere, Ignore),
-        0x20..=0xff => (Anywhere, OscPut),
+        0x20..=0x9b => (Anywhere, OscPut),
+        0x9c        => (Ground, None),
```
kchibisov: Could you explain the idea behind `0x9c        => (Ground, None)`?

src/definitions.rs
chrisduerr: Because it implements `Copy` and would cause a clippy error otherwise.

src/table.rs
chrisduerr: 0x9c is the C1 escape for the `ST` terminator.

src/definitions.rs
kchibisov: Could you also solve it here https://github.com/alacritty/vte/blob/7399615d1f94676b99defff0ebe9a385e26d7199/utf8parse/src/types.rs#L61

chrisduerr: Of course, looks like my clippy results were still cached.

Added clippy errors to all those libs, so CI should now catch that too.

PR #45, opened 2020-01-29
Bump version to 0.6.0

chrisduerr:


kchibisov:
theoretically we can also bump utf8 parse, right? It doesn't make much sense tbh, since we just added a small refactoring to it.

chrisduerr:
I thin it's not worth it tbh, since the only relevant change is the repository URL in the Cargo.toml and that redirects just fine. So I think we can just stick to what we have right now until there are actual code changes.

PR #47, opened 2020-01-31
Fix param not resetting between escapes

### LINKED ISSUE #46: DCS parameters not reset at unhook
alacritty 0.4.1

It seems parameter(s) are not reset at the end of an DCS. Some examples:

"hello" should be colored red, but isn't:
```sh
echo -e '\eP1X\e\\\e[31mhello\e[m'
```
"hello" should **not** be colored, but is:
```sh
echo -e '\eP31X\e\\\e[mhello\e[m'
```
"hello" should **not** be bold, but is:
```sh
echo -e '\eP1X\e\\\e[mhello\e[m'
```

All the above works in xterm and urxvt.

For a real-world example, see https://codeberg.org/dnkl/page, it uses the (not yet standardized) [BSU/ESU](https://gitlab.freedesktop.org/terminal-wg/specifications/merge_requests/2) DCS sequences and it is completely broken in Alacritty (do `cat <large-file> | page` and scroll).

chrisduerr:
Fixes #46.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -323,8 +323,9 @@ impl Parser {
             },
             Action::Clear => {
                 self.intermediate_idx = 0;
-                self.num_params = 0;
                 self.ignoring = false;
+                self.num_params = 0;
+                self.param = 0;
             },
             Action::BeginUtf8 => {
                 self.process_utf8(performer, byte);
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -709,6 +710,19 @@ mod tests {
         assert_eq!(dispatcher.params[0], &[i64::MAX as i64]);
     }
 
+    #[test]
+    fn reset_param_on_clear() {
+        static INPUT: &[u8] = b"\x1bP1X\x9c\x1b[31mH";
+        let mut dispatcher = CsiDispatcher::default();
```
dnkl: Isn't it a bit inconsistent to use 7-bit escapes for the DCS start and CSI, but an 8-bit escape for ST

chrisduerr: For a test it's both the same, but it's easier to understand `\x9c` than `\x1b\\` imho. I'm quite used to seeing `\x1b[` or `\x1b]`, but I just can't quite get used to `\x1b\\`, it might be because it requires `\` to be escaped?

dnkl: Good enough for me :)

Issue #38, opened 2020-01-08
Handling invalid UTF_8 bytes


sunfishcode:
I'm looking at using vte for a use case where I want to translate invalid UTF-8 bytes into Unicode replacement characters, however vte seem to silently swallow some invalid UTF-8 bytes. For example, if I feed it input consisting of the byte 0x90, it produces no events.

Would it make sense to add `Execute` rules to the `Ground` table for 0x90 and other formerly special C1 codes?

Would it make sense to introduce something like a `InvalidUtf8` action, to fill in the `Ground` table in general?

chrisduerr:
Non-utf8 8-bit C1 escapes should be passed to `execute`, so you should be able to handle C1 codes if that's your issue?

sunfishcode:
Here's a more specific testcase:
```sh
$ echo -e '\x90' > test.txt
$ target/debug/examples/parselog < test.txt
[execute] 0a
$
```
The 0x90 byte is silently dropped with no `execute` or any other action.

chrisduerr:
`\x90` is an escape introducer, which is stripped for security based on my understanding of the code.

So escapes like `\x85` will emit an `execute`, but the `DCS`(`x90`)/`CSI`(`x9b`)/`OSC`(`x9d`) 8-bit escapes are ignored.

sunfishcode:
I don't actually want to interpret C1 controls in my use case; I want to replace all non-UTF-8 bytes into replacement characters.

Right now, vte doesn't support that, either for bytes like 0x90 which are C1 controls, or bytes like 0xfd which are not. Is this a use case vte is interested in supporting?

chrisduerr:
> Is this a use case vte is interested in supporting?

I'm not sure if it's possible to support that without removing existing functionality.

Take things like the `NEL` non-utf8 8-bit C1 escape `\x85`. We trigger the `execute` function for that with this byte attached. So it's a valid escape that we propagate upstream for handling. So it's not actually invalid at all.

You could just handle `C1` escapes in your application by printing the missing glyph symbol, would that be reasonable? As far as I can tell, all that would be required then would be to make them all available appropriately.

sunfishcode:
> You could just handle C1 escapes in your application by printing the missing glyph symbol, would that be reasonable? As far as I can tell, all that would be required then would be to make them all available appropriately.

[...]

PR #58, opened 2020-06-15
Process all non-ASCII bytes with the UTF-8 parser.

sunfishcode:
The UTF-8 parser knows how to handle invalid byte sequences, so don't
preprocess the input in the main parser; just hand any non-ASCII byte to
the UTF-8 parser to handle.

This includes what were previously interpreted as 8-bit C1 control codes;
they are now interpreted as UTF-8 continuation characters.

This is the first in a series of patches which together fix https://github.com/alacritty/vte/issues/38 and a few related issues, in total enabling vte to process non-UTF-8 bytes in a manner consistent with Rust's `String::from_utf8_lossy`.

Removing support for the 8-bit C1 codes isn't necessary to achieve this, and I can revise this patch to restore support if desired, but removing them does simplify the table and the API. And the 8-bit C1 codes are rarely used and not well supported anyway. vte already doesn't appear to recognize 8-bit DCS, CSI, OSC, PM, or APC, Alacritty doesn't appear to intepret 8-bit NEL, SS2, or SS3, and [the remaining 8-bit C1 codes are rarely used](https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_control_codes_for_general_use).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+## 0.9.0
+
+- Remove 8-bit C1 support. 8-bit C1 codes are now interpreted as UTF-8
+  continuation bytes.
+
 ## 0.8.0
 
 - Remove C1 ST support in OSCs, fixing OSCs with ST in the payload
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -348,7 +348,7 @@ pub trait Perform {
     /// Draw a character to the screen and update states.
     fn print(&mut self, _: char);
 
-    /// Execute a C0 or C1 control function.
+    /// Execute a C0 control function.
     fn execute(&mut self, byte: u8);
 
     /// Invoked when a final character arrives in first part of device control string.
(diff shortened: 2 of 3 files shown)


sunfishcode:
I've now posted #60 with the full series this is part of.

The motivation for this is that from a UTF-8 perspective, 8-bit C1 control codes are continuation bytes, and if they appear alone, they form invalid UTF-8 sequences. Passing them to `execute` allows users to emit replacement characters themselves, but it's a little awkward if vte translates some invalid sequences to replacement characters and not others. It also turns out to be tricky to emit the [expected number of replacement characters](https://hsivonen.fi/broken-utf-8/) from `execute` because vte is processing some of the bytes and not others. Since vte is emitting replacement characters for some invalid sequences, and has most of the logic needed to determine how many replacement characters to emit, the simplest approach is just to have vte emit the replacement characters.

chrisduerr:
Closing in favor of #60, I'll take a look once that is in a reviewable state.

PR #61, opened 2020-07-18
Add sourcehut builds

chrisduerr:


kchibisov:
The build is failing, since Lock file from a newer version was in place, and so older version failed due to that.

PR #64, opened 2020-07-29
migrate ansi mod from alacritty_terminal

nw0:
This is from https://github.com/alacritty/alacritty/issues/1063.

I've naively moved everything over, and put `Line`, `Column`, `Rgb` in a new file, `term.rs`. The new modules, `ansi` and `term`, are behind a feature flag `ansi`, which doesn't work with `no_std` (depends on `std::io`).

As this is separate from the main crate, not filing a PR to remove from there yet.


nixpulvis:
I'd love to see what this looks like integrated with Alacritty itself.

I'm also wondering if it makes sense to move more of the functionality of `Rgb`, for example, over here. I'm not sure I would want to wrap `Rgb` in another `struct` inside Alacritty. Or do you have a plan for this?

nixpulvis:
Yea, I'm just trying to imagine how the `Rgb` `struct` should be handled between `alacritty_terminal` and `vte`. Perhaps we end up with two structs (one wrapping the other), but if there's a clean way to avoid it, that would be nice.

nw0:
> I'd really like to see this work without std too, I don't think fundamentally there are a lot of things that require std just for the dispatch part.

I believe all usages of `Write` are for replying, e.g. like "Cursor Position Report". This could be refactored into the `Handler` trait. What do you think?

> I'm just trying to imagine how the Rgb struct should be handled between alacritty_terminal and vte.

I think it makes sense for `Rgb` struct in `vte`, since we do have to handle color parsing. If we don't want serde here, we can consider a remote derive in `alacritty_terminal`?

`Line`, `Column`: it does seem a bit silly to force implementers of  `Handler` to use these newtypes. I'll remove them if there are no objections.

chrisduerr:
> I believe all usages of Write are for replying, e.g. like "Cursor Position Report". This could be refactored into the Handler trait. What do you think?

That is correct, but I don't think it makes sense to have them as part of the handler trait since they are separate things. So it should probably be a separate trait.

> I think it makes sense for Rgb struct in vte, since we do have to handle color parsing. If we don't want serde here, we can consider a remote derive in alacritty_terminal?

I've considered removing `Rgb` from vte and all associated parsing stuff, but this is really the kinda stuff why we *want* to move this to VTE. This isn't anything terminal specific, every terminal needs to behave the same here and if something is missing we want to benefit from contributions to vte in Alacritty. So really this should be part of vte. Because of that the only option is to have our Alacritty-specific derive in a wrapper type, which shouldn't be a problem.

> Line, Column: it does seem a bit silly to force implementers of Handler to use these newtypes. I'll remove them if there are no objections.

I agree, using types for this with no methods on these types just makes things confusing. Properly naming the parameters should be sufficient and easier to handle downstream. Then people can wrap it in their types anyways.

nw0:
> That is correct, but I don't think it makes sense to have them as part of the handler trait since they are separate things. So it should probably be a separate trait.

I've experimented with a trait for this. It seems that the `Handler` should be able to do whatever it wants with its "writer", rather than being constrained to `io::Write`. I've parameterised this as `trait Handler<W>`, with no constraints on `W`. Does this work?

I've removed the std dependencies, although I note that we can't use `join` (https://github.com/alacritty/vte/pull/64/files#diff-f20c875fb67cf4f8bc670d89d1d49db1R788) in 1.36.0 without the unstable `SliceConcatExt` trait. From 1.38.0, `join` is provided by a stable `SliceConcat` trait, if that's at all helpful.

chrisduerr:
> I've experimented with a trait for this. It seems that the Handler should be able to do whatever it wants with its "writer", rather than being constrained to io::Write. I've parameterised this as trait Handler<W>, with no constraints on W. Does this work?

With how traits work in Rust, that would probably mean that the handler wouldn't be able to do anything with its writer. If a trait is passed as parameter unconstrained without any methods, you can't do anything with it.

> I've removed the std dependencies, although I note that we can't use join (https://github.com/alacritty/vte/pull/64/files#diff-f20c875fb67cf4f8bc670d89d1d49db1R788) in 1.36.0 without the unstable SliceConcatExt trait (thus: updated build tests fail). From 1.38.0, join is provided by a stable SliceConcat trait, if that's at all helpful.

I don't think 1.38.0 would help, right? Since this is still using allocations like Strings and Vecs.

nw0:
> > I've experimented with a trait for this. It seems that the Handler should be able to do whatever it wants with its "writer", rather than being constrained to io::Write. I've parameterised this as trait Handler, with no constraints on W. Does this work?
> 
> With how traits work in Rust, that would probably mean that the handler wouldn't be able to do anything with its writer. If a trait is passed as parameter unconstrained without any methods, you can't do anything with it.

Actually, I'm thinking of something like `impl Handler<std::fs::File> for Term`. Parameterising on a trait would probably result in the scenario you state, yes.

---

I'm getting bogged down dealing with the `Rgb` issue at the same time as the trait one. Do you mind if I refactor this over a few PRs? I'll also work on changes to the main crate simultaneously so we're convinced the design works.

chrisduerr:
> Do you mind if I refactor this over a few PRs?

I do mind, yes. I don't think it makes sense to merge this unless there's a complete package that makes sense. Just copy-pasting stuff over isn't really beneficial.

nixpulvis:
> I'm getting bogged down dealing with the `Rgb` issue at the same time as the trait one. Do you mind if I refactor this over a few PRs?

@nw0 FWIW, it's completely possible to make a new set of branches which you merge into this branch. I'm not necessarily saying you should do that, since those PRs would be on your own fork, but it may help you personally. I know I often need a branch to keep my head on straight.

nw0:
I've straightened things out and have the following design as pushed above:

- Parameterise `Handler<W>`. This means `impl<...> Handler<W> for Term<T, W>` in the main crate -- i.e. additional parameter for `Term` et al. I'd very much like to hear about better designs, though.
- Removed serde, replace with remote derives in main crate. I had trouble with the `CursorStyle` remote deserialisation, but it doesn't seem like an issue with this design.
- Line, Column typedef'ed to `usize`.
- I do have a [working (passes tests on x86-64 Linux) implementation of the main crate as a proof-of-concept](https://github.com/nw0/alacritty/tree/ansi2).

The obvious questions:
- Structure and documentation: hidden behind a feature flag? etc.
- Remove `log`?

As regards 1.36:

chrisduerr:
> The issue is really that the trait that provides join ( rust-lang/rust#62403) was only stabilised in 1.38.0, if one has no_std.

Updating to 1.38.0 is not a problem, but even if you update you won't be able to call join here.

src/ansi.rs
```diff
@@ -0,0 +1,1607 @@
+//! ANSI Terminal Stream Parsing.
+
+extern crate alloc;
+use alloc::{borrow::ToOwned, string::String, vec::Vec};
```
chrisduerr, line 4: This would basically make this crate STD. We should try to make things allocation free.

nw0: Looks like it's mainly the window title handling code. I'll look at it over the next week or so

chrisduerr: I think it's okay to do a little less in some cases, usually handling collection from bytes into Strings is not that big of a deal.

So I think we should be able to completely remove allocations.

nw0: I've done this and I believe the only external change is the signature of `set_title`:
```rust
fn set_title(&mut self, _: Option<&str>)
```
Previously: `Option<String>`

Using ArrayVec, ArrayString -- allocation sizes are `MAX_PARAMS` or `MAX_OSC_RAW` from crate root as appropriate.

src/ansi.rs
```diff
@@ -0,0 +1,1625 @@
+//! ANSI Terminal Stream Parsing.
```
chrisduerr, line 1: The name of this file, this documentation and the feature isn't really great. That just describes all of this library basically.

src/ansi.rs
```diff
@@ -0,0 +1,1625 @@
...
+use core::str;
+
+#[cfg(feature = "no_std")]
+use arrayvec::{ArrayString, ArrayVec};
+use log::{debug, trace};
+
+use crate::{Parser, Perform};
+
+type Line = usize;
+type Column = usize;
```
chrisduerr, line 12: What are these type aliases for?

src/ansi.rs
```diff
@@ -0,0 +1,1625 @@
...
+        None
+    }
+}
+
+/// Parse colors in `rgb:r(rrr)/g(ggg)/b(bbb)` format.
+fn parse_rgb_color(color: &[u8]) -> Option<Rgb> {
+    #[cfg(feature = "no_std")]
+    let colors = str::from_utf8(color).ok()?.split('/').collect::<ArrayVec<[_; 4]>>();
+    #[cfg(not(feature = "no_std"))]
+    let colors = str::from_utf8(color).ok()?.split('/').collect::<Vec<_>>();
```
chrisduerr, line 37: Seems like this function can be easily rewritten to not require any of this Vec/ArrayVec?

src/ansi.rs
```diff
@@ -0,0 +1,1625 @@
...
+            return;
+        }
+
+        match params[0] {
+            // Set window title.
+            b"0" | b"2" => {
+                if params.len() >= 2 {
+                    #[cfg(feature = "no_std")]
+                    {
+                        let mut title = ArrayString::<[_; crate::MAX_OSC_RAW]>::new();
```
chrisduerr, line 799: I'm not familiar with ArrayString, but we have ascii as input while ArrayString most likely handles unicode. That means that the size of MAX_OSC_RAW would not be enough to store the title.

Though I see no harm in just passing params downstream and letting them handle this. Then they can decide what they want to use, no need to try and convert this into a string for them when it's trivial to do for them, but difficult to do for us.

chrisduerr:
https://github.com/alacritty/vte/pull/75

PR #65, opened 2020-08-02
Add CSI subparameter support

### LINKED ISSUE #22: Implement colon separated CSI parameters
I was looking into implementing styled underlines into alacritty, like what you might see in kitty or gnome vte based terminals - that is normal underline, double underline, curly underline (undercurl), and have it have a different color than the foreground.

Looking at the kitty implementation ([spec](https://sw.kovidgoyal.net/kitty/protocol-extensions.html#id2)), the terminal emulator has to handle not only semicolon separated parameters (`\e[0;4m`), but also colon separated parameters (`\e[4:3m`).

I'm not an expert on the standard defining how those sequences should be handled, so I was going by kitty's source code. What's being done in their codebase is not applicable to this repo, because they have the entire sequence buffered up until the command byte ('m' in this case) and then parse it, dispatching every semicolon separated parameter as it's own sequence, with special cases for color parameters. In that way, a command like `\e[0;4;58:2::186:93:0m` would be split into three function calls: handle `0m`, handle `4m`, handle `58:2::186:93:0m`.

This could be implemented in the vte crate (I think) by adding a dimension to the `params` array, and multiple calls to the Performer implementor. Not sure what the performance cost would be though.

chrisduerr:
This adds support for CSI subparameters like `\x1b[38:2:255:0:255m`,
which allows the combination of truecolor SGR commands together with
other SGR parameters like bold text, without any ambiguity.

This implements subparameters by storing them in a list together with
all other parameters and having a separate slice to indicate which
parameter is a subparameter and how long the subparameter list is. This
allows for static memory allocation and good performance while still
having the option for dynamic sizing of the parameters. Since the
subparameters are now also counted as parameters, the number of allowed
parameters has been increased from `16` to `32`.

Since the existing structures combine the handling of parameters for CSI
and DCS escape sequences, it is now also possible for DCS parameters to
have subparameters, even though that is currently never used.
Considering that DCS is rarely supported by terminal emulators, handling
these separately would likely just cause unnecessary issues. The
performance should also be better by using this existing subparam
structure rather than having two separate structures for DCS and CSI
parameters.

The only API provided for accessing the list of parameters is using an
iterator, this is intentional to make the internal structure clear and
allow for easy optimizations downstream. Since it makes little sense to
access parameters out of order, this limitation should not have any
negative effects on performance. The main drawback is that direct access
to the first parameter while ignoring all other subparameters is less
efficient, since it requires indexing a slice after iterating to the
element. However while this is often useful, it's mostly done for the
first few parameters which significantly reduces the overhead to a
negligible amount. At the same time this forces people to support
subparameters or at least consider their existence, which should make it
more difficult to implement things improperly downstream.

Fixes #22.


src/params.rs
```diff
@@ -0,0 +1,142 @@
...
+    current_subparams: u8,
+
+    /// Total number of parameters and subparameters.
+    len: usize,
+}
+
+impl Params {
+    /// Returns the number of parameters.
+    #[inline]
+    pub fn len(&self) -> usize {
```
kchibisov, line 30: I do wonder if we should use `#[inline(always)]` to ensure the inlining, however I'm not sure that it's a good idea, so maybe just `#[inline]` is fine.

src/params.rs
```diff
@@ -0,0 +1,142 @@
...
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+/// Immutable subparameter iterator.
+pub struct ParamsIter<'a> {
+    params: &'a Params,
+    index: usize,
+}
```
kchibisov, line 90: I wonder if it should be `std::iter::Peekable` iterator? I do see that we don't really need it in alacritty, but it could be useful in general or you think that there's no much value in it?

src/params.rs
chrisduerr: Inlining is a science in and of itself, though generally `#[inline(always)]` is not recommended afaik. This is also why STD has only `#[inline]` on `Vec::len` for example.

So I think if llvm has a really good reason not to inline it, sure. It should be fine.

src/params.rs
chrisduerr: You can peek any iterator by calling `peekable` on it. So no, just a generic `Iterator` should be better I think.

PR #66, opened 2020-08-27
Add LICENSE-* into vte_generate-state_changes

ignatenkobrain:
Signed-off-by: Igor Raits <i.gnatenko.brain@gmail.com>

### DIFF
--- a/vte_generate_state_changes/LICENSE-APACHE
+++ b/vte_generate_state_changes/LICENSE-APACHE
@@ -1,0 +1,176 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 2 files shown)


chrisduerr (commented):
Does it have any specific reason why you are adding these license files? The license for cargo is already part of the Cargo.toml, so the license files in the root should be sufficient.

ignatenkobrain:
Same reason I did it last time :)

When packaging crates, we take an archive from crates.io. All crates are separate units, so license should be present in all of them.

kchibisov:
It seems like Apache license is missing the actual copyright part. Also, since you're adding `MIT` here, and it wasn't there before it should be updated as well.

@chrisduerr what do you think?

chrisduerr:
> @chrisduerr what do you think?

I think the licenses should just be symlinks to the root directory. That's true for utf8parse too.

I also think that a crates.io packaging infrastructure should be able to look at the `license` field of the Cargo.toml, but that's a different story.

ghcom275 (approved):

chrisduerr:
Closing due to inactivity. If someone is interested in picking this back up, please open a new PR.

PR #69, opened 2020-11-25
Reduce the maximum value of CSI parameters

chrisduerr:
Since limits CSI parameters to be within range of `u16`, rather than
`i64`. This should effectively prevent downstream users from running
into DoS problems with excessively big escape sequence requests. An
example of a problematic escape would be `CSI Ps b` (repeat char).

According to https://vt100.net/emu/dec_ansi_parser, the smallest
possible size limit would be `16383`:

> The VT500 Programmer Information is inconsistent regarding the maximum
> value that a parameter can take. In section 4.3.3.2 of EK-VT520-RM it
> says that “any parameter greater than 9999 (decimal) is set to 9999
> (decimal)”. However, in the description of DECSR (Secure Reset), its
> parameter is allowed to range from 0 to 16383. Because individual
> control functions need to make sure that numeric parameters are within
> specific limits, the supported maximum is not critical, but it must be
> at least 16383.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+- Changed the type of CSI parameters from i64 to u16
+
 ## 0.9.0
 
 - Added CSI subparameter support; required changes can be seen in Alacritty:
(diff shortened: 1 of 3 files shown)


kchibisov (changes requested):
Looks fine, but you should fix CI build.

PR #72, opened 2021-02-21
Fix intermediate reset when going from DCS to ESC

chrisduerr:
This resolves a bug when transitioning between DCS and ESC sequences,
which would cause the intermediates of the ESC dispatch to contain data
from the DCS sequence.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -431,323 +431,141 @@ mod tests {
     ];
 
     #[derive(Default)]
-    struct OscDispatcher {
-        dispatched_osc: bool,
-        bell_terminated: bool,
-        params: Vec<Vec<u8>>,
+    struct Dispatcher {
+        dispatched: Vec<Sequence>,
     }
 
-    // All empty bodies except osc_dispatch
-    impl Perform for OscDispatcher {
+    #[derive(Debug, PartialEq, Eq)]
+    enum Sequence {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -433,178 +431,224 @@ mod tests {
...
-            self.bell_terminated = bell_terminated;
-            self.params = params.iter().map(|p| p.to_vec()).collect();
-        }
+    #[derive(Debug, PartialEq, Eq)]
+    enum Sequence {
+        Osc(Vec<Vec<u8>>, bool),
+        Csi(Vec<Vec<u16>>, Vec<u8>, bool, char),
+        Esc(Vec<u8>, bool, u8),
+        DcsHook(Vec<Vec<u16>>, Vec<u8>, bool, char),
+        DcsPut(u8),
```
nixpulvis, line 444: Would these be better served as `{}` struct variants of the enum to help make it clear what the fields mean?

chrisduerr: I've considered that, but it's test code so I didn't really care that much to be honest.

Depends on how you work with your code though I suppose. For me that makes no difference, but for some people it would probably be easier to understand.

nixpulvis: Yea, I mean it's documented in the deconstructing I suppose. I just often like to read definitions to get a sense of what's going on.

PR #74, opened 2021-05-28
Use repr(u8) on State/Action

thomcc:
Right now they'd be repr(Rust), which isn't guaranteed. If that repr changed (for example under a hypothetical `-Z repr-rust-layout=randomize`, which has been discussed), this code would fail to compile (the transmute later requires u8).

### DIFF
--- a/src/definitions.rs
+++ b/src/definitions.rs
@@ -2,6 +2,7 @@ use core::mem;
 
 #[allow(dead_code)]
 #[derive(Debug, Copy, Clone)]
+#[repr(u8)]
 pub enum State {
     Anywhere = 0,
     CsiEntry = 1,
@@ -29,6 +30,7 @@ impl Default for State {
 
 #[allow(dead_code)]
 #[derive(Debug, Clone, Copy)]
+#[repr(u8)]
 pub enum Action {
     None = 0,
     Clear = 1,


thomcc:
Build failures are unrelated clippy errors due to new clippy.

thomcc:
It's a trivial change that keeps you from relying on not-guaranteed properties...

TheEdward162:
@chrisduerr This is clearly UB as it doesn't have a stable representation ([source](https://doc.rust-lang.org/nomicon/other-reprs.html#repru-repri)) and then is used [here](https://github.com/thomcc/vte/blob/fe1022f6e6d6012652554e1964601d2827d7048f/src/definitions.rs#L59), which is C level evilness, except the layout isn't even stable.

Furthermore, this code clearly does something highly unsafe, prone to memory invariant volation, and instead of marking the function `unsafe` it just says "Bad things will happen if those invariants are violated" in the doc comment. Go back to the C corner and be ashamed.

TheEdward162:
To offer a solution with my shaming, how about reimplementing the `unpack` function in the following manner:

```rs
enum Foo {
	A = 0,
	B = 1,
	C = 2,
	D = 3,
	E = 4,
	F = 5,
	G = 6,
	H = 7,
	I = 8,
	J = 9,
	K = 10,
	L = 11,
	M = 12,
	N = 13,
	O = 14,
	P = 15
}

fn unpack(value: u8) -> Foo {
	match value & 0x0F {
		0 => Foo::A,
		1 => Foo::B,
		2 => Foo::C,
		3 => Foo::D,
		4 => Foo::E,
		5 => Foo::F,
		6 => Foo::G,
		7 => Foo::H,
		8 => Foo::I,
		9 => Foo::J,
		10 => Foo::K,
		11 => Foo::L,
		12 => Foo::M,
		13 => Foo::N,
		14 => Foo::O,
		15 => Foo::P,
		_ => unsafe { std::hint::unreachable_unchecked() }
	}
}
```

I know it's very verbose, but here the scope of unsafe doesn't reach beyond the function (i.e. to an enum definition outside the function) and is easily provable, so in my eyes it's way better. It also doesn't yell about dead code (or does and thus points at a possible error in implementation) and doesn't even require a stable repr on the enum, although since `pack` uses the numbers again the explicit values would stay.

I'm not sure about the performance implications, it's only my guess that the compiler might be able to see the pattern and hopefully do the transmute anyway, but if absolute performance isn't a requirement the unsafe can even be replaced by `unreachable!()` macro to just panic, or a panic in debug mode and unreachable_unchecked in release mode.

The verbosity could also be lightened with a macro if that's an issue, but I personally don't think it's needed, the file is short anyway.

PR #75, opened 2021-06-29
Migrate ansi mod from alacritty_terminal

alesharik:
This is continuation of #64 
Also this PR:
- adds `no_alloc` feature disable all things that need allocator
- syncs up `ansi` module with a [reference implementation](https://github.com/alacritty/alacritty/blob/master/alacritty_terminal/src/ansi.rs) in `alacritty_terminal` crate


chrisduerr:
@alesharik Out of curiosity, do you actually have an application for this personally?

src/lib.rs
```diff
@@ -34,23 +34,33 @@
...
+use std::vec::Vec;
+
 use utf8parse as utf8;
 
 mod definitions;
 mod params;
 mod table;
 
+#[cfg(feature = "ansi")]
+pub mod ansi;
```
chrisduerr, line 56:
```suggestion
#[cfg(feature = "ansi")]
pub mod ansi;

mod definitions;
mod params;
mod table;
```

Cargo.toml
```diff
@@ -15,11 +15,14 @@ edition = "2018"
...
 arrayvec = { version = "0.5.1", default-features = false, optional = true }
+log = { version = "0.4", optional = true }
 utf8parse = { version = "0.2.0", path = "utf8parse" }
 
 [features]
+ansi = ["log"]
 default = ["no_std"]
-no_std = ["arrayvec"]
+no_alloc = ["arrayvec", "no_std"]
+no_std = []
```
chrisduerr, line 25: What's the reason for changing `no_std` into `no_alloc`? Especially since this is a significant breaking change.

.builds/linux.yml
```diff
@@ -7,9 +7,15 @@ tasks:
   - stable: |
       cd vte
       $HOME/.cargo/bin/cargo +stable test
+      $HOME/.cargo/bin/cargo +stable test --features=ansi
+      $HOME/.cargo/bin/cargo +stable test --features=ansi --no-default-features
+      $HOME/.cargo/bin/cargo +stable test --features=ansi,no_alloc
```
chrisduerr, line 12: I see little point of having so many different test variations. Without ansi shouldn't make any difference when things work with it.

chrisduerr (changes requested):

Cargo.toml
alesharik: I added `no_alloc` feature to use proper structs in no_std environments with allocator. It does not replace `no_std` and should now (in commit 73e79c7) be fully backwards-compatible when `default` feature set is used. If you want, I can invert it to make it opt-in (add `alloc` feature instead)

.builds/linux.yml
alesharik: Fixed in 1b51612

Cargo.toml
chrisduerr: It should be opt-in, but I question its use. I'm not even sure if anyone is interested in running this in `no_std`.

alesharik:
@chrisduerr Hi, sorry for the delay... Yes, I'm currently using my fork of `vte` to implement basic terminal in my [hobby OS](https://github.com/alesharik/distros)

chrisduerr:
Do you make use of a custom allocator there, or are you planning to use this without allocator?

alesharik:
I'm using the crate with custom allocator

Cargo.toml
alesharik: If you can say that `osc_raw` in `VtUtf8Receiver` (no_alloc capacity is 1024), `attrs` in `attrs_from_sgr_parameters` (no_alloc capacity is 32) and `buffer` in `SyncState` (no_alloc capacity is 4Kib) very rarely overflow, then `no_alloc` is meaningless. Otherwise, we should provide the possibility to counteract these overflows.

chrisduerr:
Okay, so your usecase would be no_std + alloc then, right?

src/ansi.rs
```diff
@@ -0,0 +1,1814 @@
...
+    }
+
+    #[inline]
+    fn unhook(&mut self) {
+        match self.state.dcs {
+            Some(Dcs::SyncStart) => {
+                #[cfg(not(feature = "no_std"))]
+                {
+                    self.state.sync_state.timeout = Some(Instant::now() + SYNC_UPDATE_TIMEOUT);
+                }
```
chrisduerr, line 1023: I'd just like to point out that this completely breaks this crate when it is excluded. Just randomly leaving out arbitrary statements when they don't work with a specific feature isn't really a good way to go about things.

alesharik:
Yes, it is

chrisduerr:
An allocator feature alone might be sufficient in that case. I'm not sure what we need STD for.

src/ansi.rs
alesharik: Unfortunately, no_std does not have `Instant` struct nor require the developer to implement any means to get current time. But I can create `TimeProvider` trait which returns the current time in milliseconds as `u64` and add it to `Processor` as an argument. What do you think, is this solution suitable?

alesharik:
We need `std` just for `Instant` struct

chrisduerr:
That seems like something that could potentially be worked around.

.builds/linux.yml
```diff
@@ -6,10 +6,12 @@ tasks:
...
   - stable: |
       cd vte
-      $HOME/.cargo/bin/cargo +stable test
+      $HOME/.cargo/bin/cargo +stable test --features=ansi
+      $HOME/.cargo/bin/cargo +stable test --features=ansi,alloc
   - clippy: |
       cd vte
-      $HOME/.cargo/bin/cargo +stable clippy
+      $HOME/.cargo/bin/cargo +stable clippy --features=ansi
+      $HOME/.cargo/bin/cargo +stable clippy --features=ansi,alloc
```
chrisduerr, line 14: Would just `--all-features` work here? I think just having a general test case for all features is probably the most useful testcase. After that a 'minimal' nostd thing could be considered.

src/ansi.rs
```diff
@@ -0,0 +1,1821 @@
...
+use alloc::string::String;
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
+
+#[cfg(not(feature = "alloc"))]
+use arrayvec::ArrayVec;
+use log::{debug, trace};
+
+use crate::{Params, ParamsIter, Parser, Perform};
+use core::option::Option::Some;
```
chrisduerr, line 18: Imports should be ordered appropriately:

```suggestion
use core::convert::TryFrom;
use core::{iter, str};
use core::option::Option::Some;
use core::time::Duration;

#[cfg(feature = "alloc")]
use alloc::string::String;
#[cfg(feature = "alloc")]
use alloc::vec::Vec;

#[cfg(not(feature = "alloc"))]
use arrayvec::ArrayVec;
use log::{debug, trace};

use crate::{Params, ParamsIter, Parser, Perform};
```

src/ansi.rs
```diff
@@ -0,0 +1,1821 @@
...
+/// Maximum time before a synchronized update is aborted.
+const SYNC_UPDATE_TIMEOUT: Duration = Duration::from_millis(150);
+
+/// Maximum number of bytes read in one synchronized update (2MiB).
+#[cfg(feature = "alloc")]
+const SYNC_BUFFER_SIZE: usize = 0x20_0000;
+
+/// Maximum number of bytes read in one synchronized update (4Kib) when no_alloc feature is set.
+#[cfg(not(feature = "alloc"))]
+const SYNC_BUFFER_SIZE_NO_ALLOC: usize = 0x4000;
```
chrisduerr, line 36: I'm afraid a buffer size of 4KiB would make this feature utterly useless.

chrisduerr (changes requested):

.builds/linux.yml
alesharik: No, it won't. `--all-features` will enable `nightly` feature, which does not work on stable channel.

src/ansi.rs
alesharik: Is there any buffer size, which makes this feature work for most use cases? I don't think that this library should support _all_ use cases in no_std systems without allocator.

chrisduerr: Yeah, the existing code uses a buffer limit already. we're using 2MiB by default, I don't think it makes sense to cut this below 1 MiB.

So that's probably not going to fly.

alesharik: Ok, I disabled all `DCS` handling when `alloc` is not set

alesharik: @chrisduerr So what should I do?
- We cannot preallocate 1 MiB buffer, because it will inflate resulting binary and make this crate unsuitable for embedded systems
- We cannot use buffer smaller than 1 MiB
- We cannot disable `DCS` feature when no allocator is present

chrisduerr: My problem's just with disabling `DCS` handling in general. I have no problem with disabling the sync updates escape.

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+pub struct NullTimeProvider {}
+
+impl TimeProvider for NullTimeProvider {
+    fn now(&self) -> u128 {
+        0
+    }
+}
+
+#[derive(Debug)]
+#[cfg(feature = "alloc")]
```
chrisduerr, line 152:
```suggestion
#[cfg(feature = "alloc")]
#[derive(Debug)]
```

Logically I think it makes way more sense to have the `cfg` attribute before any other macro. So I think it would be good to change this everywhere so that the scope under which the struct is defined is the first thing people read.

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+    sync_state: SyncState<T>,
+
+    _time_provider: PhantomData<T>,
+}
+
+/// Current timestamp provider
+pub trait TimeProvider {
+    /// Get current timestamp from unix epoch in milliseconds
+    fn now(&self) -> u128;
+}
```
chrisduerr, line 139: All this `TimeProvider` stuff just seems pointlessly complicated.

Is it really worth it to try and abstract over STD at that point? It probably makes more sense to exclude the sync updates from anything but STD+alloc.

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+    pub fn new() -> Processor<NullTimeProvider> {
+        Processor {
+            parser: Parser::new(),
+            state: ProcessorState {
+                preceding_char: None,
+                _time_provider: PhantomData::<NullTimeProvider>::default(),
+            },
+        }
+    }
+}
```
chrisduerr, line 215: I don't think providing these is particularly beneficial. Without alloc a time provider shouldn't even be necessary in the first place.

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+            } else {
+                self.advance_sync(handler, writer, byte)
+            }
+        }
+        #[cfg(not(feature = "alloc"))]
+        {
+            let mut performer = Performer::new(&mut self.state, handler, writer);
+            self.parser.advance(&mut performer, byte);
+        }
+    }
```
chrisduerr, line 250: What's the point of having separate blocks inside the method, when the entire method is different for each feature anyway?

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+            self.set_title_utf(params.map(|params| {
+                params[1..]
+                    .iter()
+                    .flat_map(|x| str::from_utf8(x))
+                    .collect::<Vec<&str>>()
+                    .join(";")
+                    .trim()
+                    .to_owned()
+            }))
+        }
```
chrisduerr, line 389: What exactly is the point of this? Having separate `set_title` functions just makes things complicated and easy to screw up. Performing some string manipulation of params in the `Handler` is completely inappropriate.

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+                self.state.sync_state.timeout = Some(
+                    self.state.sync_state.time_provider.now() + SYNC_UPDATE_TIMEOUT.as_millis(),
+                );
+                self.state.sync_state.in_sync = true;
+            },
+            Some(Dcs::SyncEnd) => (),
+            _ => debug!("[unhandled unhook]"),
+        }
+        #[cfg(not(feature = "alloc"))]
+        debug!("[unhandled unhook]")
```
chrisduerr, line 1063: Entire method scoped behind cfgs again *inside* the method itself.

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+            _ => unhandled!(),
+        }
+    }
+}
+
+fn attrs_from_sgr_parameters(
+    params: &mut ParamsIter<'_>,
+) -> impl IntoIterator<Item = Option<Attr>> {
+    #[cfg(not(feature = "alloc"))]
+    let mut attrs = ArrayVec::<[_; crate::params::MAX_PARAMS]>::new();
```
chrisduerr, line 1433: No fully qualified paths please. Segments should be imported as appropriate at the beginning of the file.

src/ansi.rs
```diff
@@ -0,0 +1,1849 @@
...
+            [101] => Some(Attr::Background(Color::Named(NamedColor::BrightRed))),
+            [102] => Some(Attr::Background(Color::Named(NamedColor::BrightGreen))),
+            [103] => Some(Attr::Background(Color::Named(NamedColor::BrightYellow))),
+            [104] => Some(Attr::Background(Color::Named(NamedColor::BrightBlue))),
+            [105] => Some(Attr::Background(Color::Named(NamedColor::BrightMagenta))),
+            [106] => Some(Attr::Background(Color::Named(NamedColor::BrightCyan))),
+            [107] => Some(Attr::Background(Color::Named(NamedColor::BrightWhite))),
+            _ => None,
+        };
+        attrs.push(attr);
```
chrisduerr, line 1516: This will panic without any warning should the attrs ever exceed MAX_PARAMS. There is no guarantee that this will always be the case in the future.

chrisduerr (changes requested):

src/ansi.rs
alesharik: I agree with you, but how do I remove `TimeProvider` generic parameter from `impl Perform for Performer` in no-alloc environment?

src/ansi.rs
alesharik: What should I do? Print warning about excess attrs and ignore them?

src/ansi.rs
alesharik: Hi, sorry for the delay. Yes, I tried to disable sync updates only, but I can't see a clear boundary between `DCS` and sync updates handling, so I did that I could.

src/ansi.rs
chrisduerr: Shouldn't be necessary to print a warning, but we should probably make sure it doesn't panic.

src/ansi.rs
```diff
@@ -1419,6 +1419,17 @@ fn attrs_from_sgr_parameters(
...
     let mut attrs = Vec::with_capacity(params.size_hint().0);
     while let Some(param) = params.next() {
+        #[cfg(not(feature = "alloc"))]
+        {
+            if attrs.len() == MAX_PARAMS {
+                warn!(
+                    "SGR parameter count is more than max terminal attributes count ({})!\
+                     Skipping exceeding parameters",
+                    MAX_PARAMS
+                );
```
chrisduerr, line 1429: This is not an error that makes sense logging to the user.

Andy-Python-Programmer:
Any updates on this?

chrisduerr:
> Is this still being worked on?

I doubt it.

PR #78, opened 2022-05-17
Fix filled params list ending with subparam

### LINKED ISSUE #77: Infinite loop in ParamsIter
I've been doing some fuzzing of vte, and discovered that the input `b"\x1bP;;;:::::;;:::::;;;;;;;;;;;;;0;;::;;p\x1b"` causes ParamsIter to loop forever.

I'm trying to get my head around the escape code format enough to troubleshoot more usefully, but maybe you can figure out what's going on before I can.

chrisduerr:
When the params list for the CSI/DCS escapes is filled with all 32
parameters but ends in a subparameter, it would not properly stage the
length of the added subparameters causing the param iterator to get
stuck in place.

To ensure we always update the subparameter length even when no
parameter is staged after it, the length of subparameters is now updated
immediately while the subparameters itself are added.

Fixes #77.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 ## Unreleased
 
 - Minimum rust version has been bumped to 1.56.0
+- Fixed infinite loop in `Params` iterator when 32nd parameter is a subparameter
 
 ## 0.10.1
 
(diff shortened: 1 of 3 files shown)


kchibisov (changes requested):
Should add a changelog entry wrt that fix, otherwise looks good.

kchibisov (approved):

PR #83, opened 2023-03-07
Add trivial derives to `utf8parser::Parser`

epage:
I am using this to validate output sent to `std::io::Write::write` but the entire validated output might not be written.  To workaround this, I need to snapshot my validation state, attempt the `write`, and then replay my validation up-to what was written.  This requires the data to be `clone`able to do so but `Parser` isn't.

Much like `std::ops::Range`, we likely don't want this to be `Copy` as that makes it too easy to get mixed up on what state you are using but `Clone` should be explicit enough to be safe.

In addition to `Clone`, I've also added `Debug`, `PartialEq`, and `Eq`

### DIFF
--- a/utf8parse/src/lib.rs
+++ b/utf8parse/src/lib.rs
@@ -25,7 +25,7 @@ pub trait Receiver {
 /// A parser for Utf8 Characters
 ///
 /// Repeatedly call `advance` with bytes to emit Utf8 characters
-#[derive(Default)]
+#[derive(Clone, Default)]
 pub struct Parser {
     point: u32,
     state: State,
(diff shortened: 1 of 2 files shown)


chrisduerr (changes requested):
Do you mind looking through https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits and adding all trivial derives that could be implemented without harm?

I'd imagine if `Clone` is needed, something like `PartialEq` might be useful too. That said I believe `State` doesn't implement anything other than `Debug`, so we can probably only add `Debug` or add everything to the `State` enum too (which I'm fine with, but don't want to make a requirement for this PR).

epage:
Done

chrisduerr (approved):

chrisduerr:
If you need a release for this, please let me know. Considering the slow pace of VTE, this will likely take a while before being released naturally.

PR #85, opened 2023-04-23
migrate `ansi` from `alacritty_terminal`

Andy-Python-Programmer:
This pull request is a continuation of https://github.com/alacritty/vte/pull/64 rather than https://github.com/alacritty/vte/pull/75 (which additionally added `no-alloc` support). If required, `no-alloc` support can be added in a follow up pull request.

## Unresolved Questions
* Should the `log` crate be used? It should be fine to ignore stuff that is unhanded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+- Minimum rust version has been bumped to 1.62.1
+- Support for ANSI terminal stream parsing under the `ansi` feature.
+- Addition of the `serde` feature which additionally derives `Serialize` and `Deserialize` 
+  for the types provided in the `ansi` module.
+
 ## 0.11.0
 
 - Minimum rust version has been bumped to 1.56.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ CHANGELOG
 
 - Minimum rust version has been bumped to 1.62.1
 - Support for ANSI terminal stream parsing under the `ansi` feature.
-- Addition of the `serde` feature which additionally derives `Serialize` and `Deserialize` 
+- Addition of the `serde` feature which derives `Serialize` and `Deserialize`
   for the types provided in the `ansi` module.
 
 ## 0.11.0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,9 +18,11 @@ vte_generate_state_changes = { version = "0.1.0", path = "vte_generate_state_cha
 arrayvec = { version = "0.7.2", default-features = false, optional = true }
 log = { version = "0.4.17", optional = true }
 utf8parse = { version = "0.2.0", path = "utf8parse" }
+serde = { version = "1.0.160", features = ["derive"], optional = true }
 
 [features]
 ansi = ["log"]
+ansi-serde = ["serde"]
 default = ["no_std"]
 no_std = ["arrayvec"]
 nightly = ["utf8parse/nightly"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ rust-version = "1.56.0"
 [dependencies]
 vte_generate_state_changes = { version = "0.1.0", path = "vte_generate_state_changes" }
 arrayvec = { version = "0.7.2", default-features = false, optional = true }
-log = { version = "0.4", optional = true }
+log = { version = "0.4.17", optional = true }
 utf8parse = { version = "0.2.0", path = "utf8parse" }
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ serde = { version = "1.0.160", features = ["derive"], optional = true }
 
 [features]
 ansi = ["log"]
-ansi-serde = ["serde"]
+serde = ["dep:serde"]
 default = ["no_std"]
 no_std = ["arrayvec"]
 nightly = ["utf8parse/nightly"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ serde = { version = "1.0.160", features = ["derive"], optional = true }
 ansi = ["log"]
 serde = ["dep:serde"]
 default = ["no_std"]
+floats = []
 no_std = ["arrayvec"]
 nightly = ["utf8parse/nightly"]
 
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -282,8 +282,17 @@ enum Dcs {
 }
 
 /// The processor wraps a `crate::Parser` to ultimately call methods on a Handler.
+#[cfg(not(feature = "no_std"))]
 #[derive(Default)]
 pub struct Processor<T: SyncHandler = DefaultSyncHandler> {
+    state: ProcessorState<T>,
+    parser: crate::Parser,
+}
+
+/// The processor wraps a `crate::Parser` to ultimately call methods on a Handler.
+#[cfg(feature = "no_std")]
+#[derive(Default)]
+pub struct Processor<T: SyncHandler> {
     state: ProcessorState<T>,
     parser: crate::Parser,
 }
@@ -406,12 +415,6 @@ impl<'a, H: Handler + 'a, T: SyncHandler> Performer<'a, H, T> {
 }
 
 #[cfg(not(feature = "no_std"))]
-pub type DefaultSyncHandler = StdSyncHandler;
-
-#[cfg(feature = "no_std")]
-pub type DefaultSyncHandler = NullTimeProvider;
-
-#[cfg(not(feature = "no_std"))]
 #[derive(Default)]
 pub struct StdSyncHandler {
     timeout: Option<std::time::Instant>,
@@ -436,20 +439,6 @@ impl SyncHandler for StdSyncHandler {
     #[inline]
     fn pending_timeout(&self) -> bool {
         self.timeout.is_some()
-    }
-}
-
-/// Time provider which always returns 0.
-#[derive(Default)]
-pub struct NullTimeProvider;
-
-impl SyncHandler for NullTimeProvider {
-    #[inline]
-    fn update_timeout(&mut self, _: Option<Duration>) {}
-
-    #[inline]
-    fn pending_timeout(&self) -> bool {
-        false
     }
 }
 
@@ -1747,6 +1736,21 @@ pub mod C0 {
 mod tests {
     use super::*;
 
+    #[derive(Default)]
+    pub struct TestSyncHandler;
+
+    impl SyncHandler for TestSyncHandler {
+        #[inline]
+        fn update_timeout(&mut self, _: Option<Duration>) {
+            unreachable!()
+        }
+
+        #[inline]
+        fn pending_timeout(&self) -> bool {
+            false
+        }
+    }
+
     struct MockHandler {
         index: CharsetIndex,
         charset: StandardCharset,
@@ -1804,7 +1808,7 @@ mod tests {
     fn parse_control_attribute() {
         static BYTES: &[u8] = &[0x1b, b'[', b'1', b'm'];
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in BYTES {
@@ -1818,7 +1822,7 @@ mod tests {
     fn parse_terminal_identity_csi() {
         let bytes: &[u8] = &[0x1b, b'[', b'1', b'c'];
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in bytes {
@@ -1850,7 +1854,7 @@ mod tests {
     fn parse_terminal_identity_esc() {
         let bytes: &[u8] = &[0x1b, b'Z'];
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in bytes {
@@ -1862,7 +1866,7 @@ mod tests {
 
         let bytes: &[u8] = &[0x1b, b'#', b'Z'];
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in bytes {
@@ -1880,7 +1884,7 @@ mod tests {
             b'2', b'5', b'5', b'm',
         ];
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in BYTES {
@@ -1912,7 +1916,7 @@ mod tests {
         ];
 
         let mut handler = MockHandler::default();
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
 
         for byte in BYTES {
             parser.advance(&mut handler, *byte);
@@ -1922,7 +1926,7 @@ mod tests {
     #[test]
     fn parse_designate_g0_as_line_drawing() {
         static BYTES: &[u8] = &[0x1b, b'(', b'0'];
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in BYTES {
@@ -1936,7 +1940,7 @@ mod tests {
     #[test]
     fn parse_designate_g1_as_line_drawing_and_invoke() {
         static BYTES: &[u8] = &[0x1b, b')', b'0', 0x0e];
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in &BYTES[..3] {
@@ -1999,7 +2003,7 @@ mod tests {
     fn parse_osc4_set_color() {
         let bytes: &[u8] = b"\x1b]4;0;#fff\x1b\\";
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in bytes {
@@ -2013,7 +2017,7 @@ mod tests {
     fn parse_osc104_reset_color() {
         let bytes: &[u8] = b"\x1b]104;1;\x1b\\";
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in bytes {
@@ -2027,7 +2031,7 @@ mod tests {
     fn parse_osc104_reset_all_colors() {
         let bytes: &[u8] = b"\x1b]104;\x1b\\";
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in bytes {
@@ -2042,7 +2046,7 @@ mod tests {
     fn parse_osc104_reset_all_colors_no_semicolon() {
         let bytes: &[u8] = b"\x1b]104\x1b\\";
 
-        let mut parser: Processor = Processor::new();
+        let mut parser = Processor::<TestSyncHandler>::new();
         let mut handler = MockHandler::default();
 
         for byte in bytes {
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -1,5 +1,7 @@
 //! ANSI Terminal Stream Parsing.
 
+extern crate alloc;
+
 use alloc::borrow::ToOwned;
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -455,14 +455,21 @@ impl Timeout for StdSyncHandler {
     }
 }
 
+/// Provides an interface for common timer functionality (i.e.
+/// creating timeouts and checking if it has been expired).
+///
+/// This is internally used by the [`Processor`] to handle synchronized
+/// updates.
 pub trait Timeout: Default {
     /// Sets the timeout for the next synchronized update.
     ///
-    /// The `duration` parameter specifies the duration of the timeout. Once the timeout has elapsed, the synchronized update can be performed.
+    /// The `duration` parameter specifies the duration of the timeout. Once the
+    /// specified duration has elapsed, the synchronized update rotuine can be
+    /// performed.
     fn set_timeout(&mut self, duration: Duration);
     /// Clear the current timeout.
     fn clear_timeout(&mut self);
-    /// Returns whether a timeout is currently pending.
+    /// Returns whether a timeout is currently active and has not yet expired.
     fn pending_timeout(&self) -> bool;
 }
 
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -20,7 +20,6 @@ use core::time::Duration;
 use core::{iter, str};
 
 use log::{debug, trace};
-
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -456,9 +456,9 @@ impl Timeout for StdSyncHandler {
 }
 
 pub trait Timeout: Default {
-    /// Sets the timeout for the next synchronized update. The `duration` parameter
-    /// specifies the duration of the timeout. Once the timeout has elapsed, the
-    /// synchronized update can be performed.
+    /// Sets the timeout for the next synchronized update.
+    ///
+    /// The `duration` parameter specifies the duration of the timeout. Once the timeout has elapsed, the synchronized update can be performed.
     fn set_timeout(&mut self, duration: Duration);
     /// Clear the current timeout.
     fn clear_timeout(&mut self);
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -319,11 +319,11 @@ impl<T: SyncHandler> Processor<T> {
     where
         H: Handler,
     {
-        if !self.state.sync_state.handler.pending_timeout() {
+        if self.state.sync_state.handler.pending_timeout() {
+            self.advance_sync(handler, byte);
+        } else {
             let mut performer = Performer::new(&mut self.state, handler);
             self.parser.advance(&mut performer, byte);
-        } else {
-            self.advance_sync(handler, byte);
         }
     }
 
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -22,6 +22,9 @@ use core::{iter, str};
 #[cfg(feature = "floats")]
 use core::ops::Mul;
 
+#[cfg(not(feature = "no_std"))]
+use std::time::Instant;
+
 use log::{debug, trace};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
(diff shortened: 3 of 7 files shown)


chrisduerr:
> Should the log crate be used? It should be fine to ignore stuff that is unhanded.

Yes. Not logging is not an option.

Andy-Python-Programmer:
Also should the sync updates be excluded if its `no_std` since `Instant` will not be available?

chrisduerr:
> Also should the sync updates be excluded if its no_std since Instant will not be available? Or should the timer functionality be provided by TimeProvider as with the previous pull request.

Not providing certain features isn't really an option (imo).

src/ansi.rs
```diff
@@ -0,0 +1,1913 @@
...
+pub type DefaultTimeProvider = StdTimeProvider;
+
+#[cfg(not(feature = "no_std"))]
+#[derive(Default)]
+pub struct StdTimeProvider;
+
+#[cfg(not(feature = "no_std"))]
+impl TimeProvider for StdTimeProvider {
+    fn now(&self) -> Duration {
+        todo!()
```
Andy-Python-Programmer, line 31: How should this be handled? Since AFAIK `Instant` is the only way to get monotonic time from the standard library.

Andy-Python-Programmer:
Another way we could approach this is by adding a `set_timeout(&self, _: Duration)` function on `Handler`, which will make it much more portable since now we do not have to handle all of the `Instant` and `TimeProvider` mess; making it less complex.

chrisduerr:
> Another way we could approach sync updates will be by adding a update_timeout(&self, _: Duration) function on Handler, which will make it much more portable since now we do not have to handle all of the Instant and SyncHandler mess; making it less complex.

I'm not sure what you mean? The whole point of an abstraction would be that we handle all that "mess" for implementations. People shouldn't have to know how sync updates work to support them.

Andy-Python-Programmer:
Then is the current implementation alright?

Cargo.toml
```diff
@@ -16,9 +16,11 @@ rust-version = "1.56.0"
 [dependencies]
 vte_generate_state_changes = { version = "0.1.0", path = "vte_generate_state_changes" }
 arrayvec = { version = "0.7.2", default-features = false, optional = true }
+log = { version = "0.4", optional = true }
```
chrisduerr, line 19:
```suggestion
log = { version = "0.4.17", optional = true }
```

src/ansi.rs
```diff
@@ -0,0 +1,1912 @@
...
+
+        let expected: Vec<usize> = (0..256).collect();
+        assert_eq!(handler.reset_colors, expected);
+    }
+
+    #[test]
+    fn parse_osc104_reset_all_colors_no_semicolon() {
+        let bytes: &[u8] = b"\x1b]104\x1b\\";
+
+        let mut parser: Processor = Processor::new();
```
chrisduerr, line 1902: Why did you explicitly specify the type here everywhere?

src/ansi.rs
```diff
@@ -0,0 +1,1912 @@
...
+                    self.handler.set_hyperlink(None);
+                    return;
+                }
+
+                // Link parameters are in format of `key1=value1:key2=value2`. Currently only key
+                // `id` is defined.
+                let id = link_params
+                    .split(|&b| b == b':')
+                    .find_map(|kv| kv.strip_prefix(b"id="))
+                    .and_then(|kv| str::from_utf8(kv).ok().map(|e| e.to_owned()));
```
chrisduerr, line 1090: I'd assume .cloned should work here?

src/ansi.rs
```diff
@@ -0,0 +1,1912 @@
...
+#[derive(Default)]
+pub struct NullTimeProvider;
+
+impl SyncHandler for NullTimeProvider {
+    fn update_timeout(&mut self, _: Option<Duration>) {}
+
+    fn pending_timeout(&self) -> bool {
+        false
+    }
+}
```
chrisduerr, line 56: What exactly is this supposed to be? Are you suggesting to just break everything on no_std without users being able to know what's going on?

src/ansi.rs
```diff
@@ -0,0 +1,1912 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Color {
+    Named(NamedColor),
+    Spec(Rgb),
+    Indexed(u8),
+}
```
chrisduerr, line 812: I don't think removing the Serialize/Deserialize flags here is a good idea. I believe this is required for Alacritty's tests? Have you checked or did you just remove it? This module should obviously integrate with Alacritty.

src/ansi.rs
```diff
@@ -0,0 +1,1912 @@
...
+
+    /// Cursor is a vertical bar `⎸`.
+    Beam,
+
+    /// Cursor is a box like `☐`.
+    HollowBlock,
+
+    /// Invisible cursor.
+    Hidden,
+}
```
chrisduerr, line 543: Config deserialization here is a necessary part for Alacritty.

src/ansi.rs
```diff
@@ -0,0 +1,1912 @@
...
+        }
+
+        // Resetting state after processing makes sure we don't interpret buffered sync escapes.
+        self.state.sync_state.buffer.clear();
+        self.state.sync_state.handler.update_timeout(None);
+    }
+
+    /// Number of bytes in the synchronization buffer.
+    #[inline]
+    pub fn sync_bytes_count(&self) -> usize {
```
chrisduerr, line 248: Did you just *remove* the `sync_timeout` method without any replacement?!

chrisduerr (changes requested):

src/ansi.rs
Andy-Python-Programmer: > Why did you explicitly specify the type here everywhere?

`Processor` is defined as `pub struct Processor<T: SyncHandler = DefaultSyncHandler> {` so when you just do `Processor::new()` it evaluates to something like `<Processor<?T>>` where the compiler has to infer `T`. Whereas if you specify the type, the default for the generic will be used. An alternative would be to do `<Processor>::new()` or we could manually create the default implementation for `Processor` depending on if we have the standard library available.

src/ansi.rs
Andy-Python-Programmer: `cloned()` will not work because it will return a `str` (not `Sized` so would'nt compile anyways) while we are trying to get a `String`.

src/ansi.rs
Andy-Python-Programmer: I could add the `Serialize` and `Deserialize` derives behind a `ansi-serde` feature flag.

src/ansi.rs
Andy-Python-Programmer: https://github.com/alacritty/vte/pull/85/commits/bd40f34302bc5617266de43ec4a4567869540a1a

src/ansi.rs
Andy-Python-Programmer: Alternative would be to not have a default sync handler for `no_std`, which should be fine and would force the user to create their own handler.

src/ansi.rs
Andy-Python-Programmer: https://github.com/alacritty/vte/pull/85/commits/f086eb7137b8191ddb4597a75939066968c25334

src/ansi.rs
chrisduerr: The feature flag should be `serde`, not `ansi-serde`.

src/ansi.rs
chrisduerr: And how would you suggest a `no_std` implementation to look like?

src/ansi.rs
Andy-Python-Programmer: > The feature flag should be `serde`, not `ansi-serde`.

Fixed in https://github.com/alacritty/vte/pull/85/commits/76440a7b375f43bb7c066b940cdc552c692ec6cd.

src/ansi.rs
kchibisov: It would be nice to have an alacritty PR along the way.

Andy-Python-Programmer: > And how would you suggest a `no_std` implementation to look like?

With `no_std` it really depends on where you intend to use the crate. For example, if you intend of using the crate in a kernel, then the implementation would differ from a case where you are using the `libc` to get the timer functionality. 

In any case, it will look very similar to the default implementation of the sync handler that is currently provided for `std`. Here is roughly what the sync handler would look like by using the `libc`:
```rust
struct MySyncHandler {
     timeout: Option<libc::timespec>,
}

impl vte::SyncHandler for MySyncHandler {
    fn update_timeout(&mut self, duration: Option<Duration>) {
         let mut timespec = core::mem::zeroed();
         let now = cvt(unsafe { libc::clock_gettime(CLOCK_MONOTONIC, &mut timespec) }).unwrap();

         self.timeout = duration.map(|e| timespec);
    }

    #[inline]
    fn pending_timeout(&self) -> bool {
         self.timeout.is_some()
    }
}
```

src/lib.rs
```diff
@@ -34,6 +34,8 @@
 #![cfg_attr(all(feature = "nightly", test), feature(test))]
 #![cfg_attr(feature = "no_std", no_std)]
 
+extern crate alloc;
+
```
kchibisov, line 38: We should use alloc only for ansi, and the rest should be able to be used without alloc, so alloc should be a feature.

src/lib.rs
```diff
@@ -44,6 +46,8 @@ mod definitions;
 mod params;
 mod table;
 
+#[cfg(feature = "ansi")]
```
kchibisov, line 49: The ansi should enable alloc.

src/ansi.rs
```diff
@@ -0,0 +1,1933 @@
...
+#[derive(Default)]
+pub struct NullTimeProvider;
+
+impl SyncHandler for NullTimeProvider {
+    fn update_timeout(&mut self, _: Option<Duration>) {}
+
+    fn pending_timeout(&self) -> bool {
+        false
+    }
+}
```
kchibisov, line 69: We use master layout for definitions, in a way that definition comes after the use, with some priorities. `const` stuff should be above the rest, and the structure's will move to the bottom.

kchibisov (changes requested):
The patch against alacritty should be opened to help handling the design, since it's the main consumer of such code.

src/ansi.rs
```diff
@@ -0,0 +1,1933 @@
+//! ANSI Terminal Stream Parsing.
```
kchibisov, line 1: The `SPDX` header should be used with `APACHE` license, because the `alacritty_terminal` is APACHE only.

src/lib.rs
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/9f3c6850adc997df324fe79b76ac624e6c6b6e29.

src/lib.rs
Andy-Python-Programmer: https://github.com/alacritty/vte/pull/85/commits/9f3c6850adc997df324fe79b76ac624e6c6b6e29

src/ansi.rs
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/b2d518d87d94512bcae22d7bd3a8c07b945bf7a2.

src/ansi.rs
Andy-Python-Programmer: Do the `impl`s be after the structure definition as-well? If I understood this correctly, the correct layout would be type aliases, structure definition and implementation pair and at the end will be the trait definition?

src/ansi.rs
```diff
@@ -0,0 +1,1941 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// This module was originally part of the `alacritty_terminal` crate, which is
+// licensed under the Apache License, Version 2.0 and is part of the Alacritty
+// project (https://github.com/alacritty/alacritty).
```
chrisduerr, line 5: I don't think this is necessary considering both repositories are part of the same project and both are licensed under MIT+Apache.

In its current state this is more harmful than helpful, since using it as Apache only would not technically allow usage of VTE under MIT anymore. And I don't think it's worth specifying the multi-license.

src/ansi.rs
kchibisov:
```rust
impl Other {
    fn other_fn() -> Foo {
    }
}

struct Foo {
}
impl Foo {
}
impl Drop for Foo {
}

impl OtherTraits for Foo {
}
```

Andy-Python-Programmer:
> The patch against alacritty should be opened to help handling the design, since it's the main consumer of such code.

https://github.com/alacritty/alacritty/pull/6896

src/ansi.rs
Andy-Python-Programmer: fixed in https://github.com/alacritty/vte/pull/85/commits/994c055115c4f990f759737cf14bf8f37b213131

src/ansi.rs
```diff
@@ -0,0 +1,1943 @@
...
+pub struct Hyperlink {
+    /// Identifier for the given hyperlink.
+    pub id: Option<String>,
+    /// Resource identifier of the hyperlink.
+    pub uri: String,
+}
+
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Default)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Rgb {
```
Andy-Python-Programmer, line 37: The `Rgb` struct in `alacritty_terminal` provides additional functions (`luminance` and `contrast`) and implements additional traits (`Mul<f32>` and `Add<Rgb>`).

In the initial version of making Alacritty use VTE ANSI parser (https://github.com/alacritty/alacritty/pull/6896), I just convert between the Alacritty and VTE `Rgb` structs. By adding those functions in VTE would allow us to get rid of the type from `alacritty_terminal` but will still require us to have a `RgbShim` type so that we can implement the custom deserialize for ref tests and serde replace where serialization is required.

Which approach would you guys suggest to take?

chrisduerr: Why not just move the functions to VTE? If Alacritty requires it, there's a good chance other emulators might too.

Andy-Python-Programmer: I have moved that functionality to VTE (https://github.com/alacritty/vte/pull/85/commits/95132a6ba9b23228d2b5c0511609c1b59be97dba and https://github.com/alacritty/vte/pull/85/commits/187cdc8a3bb7d12794bb918ee8acad23e236a3e7). I have also updated Alacritty to make use of the VTE provided RGB type under the hood (https://github.com/alacritty/alacritty/pull/6896/commits/37c22c800187cf78551a2756b5facb3f940fbb0d).

src/ansi.rs
Andy-Python-Programmer: I modified the `Processor` type in https://github.com/alacritty/vte/pull/85/commits/c70bc9645553c8c2238f411327e8532ff83234b2 to require a sync handler if it is being used in a `no_std` environment. Otherwise, the default `StdSyncHandler` is used. This change should alleviate the worry of unexpected issues arising in `no_std` environments without users being able to identify the cause.

Andy-Python-Programmer:
I believe I have addressed all of the requested changes in the reviews.

Cargo.toml
```diff
@@ -16,10 +16,15 @@ rust-version = "1.56.0"
...
 arrayvec = { version = "0.7.2", default-features = false, optional = true }
+log = { version = "0.4.17", optional = true }
 utf8parse = { version = "0.2.0", path = "utf8parse" }
+serde = { version = "1.0.160", features = ["derive"], optional = true }
 
 [features]
+ansi = ["log"]
+serde = ["dep:serde"]
 default = ["no_std"]
+floats = []
```
chrisduerr, line 27: Why is this behind a floats feature?

src/ansi.rs
```diff
@@ -0,0 +1,2062 @@
...
+const SYNC_BUFFER_SIZE: usize = 0x20_0000;
+
+/// Number of bytes in the synchronized update DCS sequence before the passthrough parameters.
+const SYNC_ESCAPE_START_LEN: usize = 5;
+
+/// Start of the DCS sequence for beginning synchronized updates.
+const SYNC_START_ESCAPE_START: [u8; SYNC_ESCAPE_START_LEN] = [b'\x1b', b'P', b'=', b'1', b's'];
+
+/// Start of the DCS sequence for terminating synchronized updates.
+const SYNC_END_ESCAPE_START: [u8; SYNC_ESCAPE_START_LEN] = [b'\x1b', b'P', b'=', b'2', b's'];
```
chrisduerr, line 171: Constants should be at the top of the file, not somewhere in the middle.

src/ansi.rs
```diff
@@ -0,0 +1,2062 @@
...
+            (self_luminance, other_luminance)
+        };
+
+        (lighter + 0.05) / (darker + 0.05)
+    }
+}
+
+// A multiply function for Rgb, as the default dim is just *2/3.
+#[cfg(feature = "floats")]
+impl core::ops::Mul<f32> for Rgb {
```
chrisduerr, line 87: Do not use full paths, but instead add imports to the top of the file.

src/ansi.rs
```diff
@@ -0,0 +1,2062 @@
...
+use core::fmt::{self, Display, Formatter, Write};
+use core::ops::{Add, Sub};
+use core::str::FromStr;
+use core::time::Duration;
+use core::{iter, str};
+
+use log::{debug, trace};
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
```
chrisduerr, line 25:
```suggestion
use log::{debug, trace};
#[cfg(feature = "serde")]
use serde::{Deserialize, Serialize};
```

chrisduerr (changes requested):

Cargo.toml
Andy-Python-Programmer: Anything that necessitates the use of floating point should be made optional. This is because if the crate is used use in a kernel (which is my intended use), floats are not usually available :)

src/ansi.rs
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/d9f17f1fe607a08bfea52ccbec26106093a08b0d.

src/ansi.rs
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/d9f17f1fe607a08bfea52ccbec26106093a08b0d.

src/ansi.rs
```diff
@@ -0,0 +1,2064 @@
...
+        self.timeout = duration.map(|e| Instant::now() + e);
+    }
+
+    #[inline]
+    fn pending_timeout(&self) -> bool {
+        self.timeout.is_some()
+    }
+}
+
+pub trait SyncHandler: Default {
```
chrisduerr, line 450: This trait name is honestly kinda bad because it doesn't at all do what the name says it does. There is no synchronized update handling here.

The trait itself isn't even necessarily related to synchronized updates. You could use the same trait for any timeout really.

I think just `Timeout` or something like that would be better.

Then we should also rename it wherever it's stored. Storing something as "handler" is about as useless as it gets as far as variable names go.

chrisduerr (changes requested):
This PR seems to be drifting towards completion, so I'd appreciate it if you could fix the CI issues just to avoid having to go another review round when those fixes introduce new issues.

Andy-Python-Programmer:
Hello, @chrisduerr. I have resolved the clippy errors in the CI. However, the CI is currently failing due to the unavailability of the `dep:` prefix in `1-56-0`. To resolve this we could: either upgrade the minimum supported Rust version (MSRV) or rename the serde feature to a different name to avoid conflicts with the dependency.

Based on your earlier suggestion to rename the `ansi-serde` feature to `serde`, I presume that upgrading the MSRV would be the appropriate solution.

Andy-Python-Programmer:
Updated the CI Rust version to `1.62.1` and now CI is green ✅

src/ansi.rs
```diff
@@ -0,0 +1,2064 @@
...
+    #[inline]
+    pub fn sync_timeout(&self) -> Option<std::time::Instant> {
+        self.timeout
+    }
+}
+
+#[cfg(not(feature = "no_std"))]
+impl SyncHandler for StdSyncHandler {
+    fn update_timeout(&mut self, duration: Option<Duration>) {
+        use std::time::Instant;
```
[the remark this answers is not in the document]
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/34d228cfd0cf7d9ef9fb5d5f0203323c6f3f76ab.

src/ansi.rs
```diff
@@ -0,0 +1,2064 @@
...
+    #[inline]
+    pub fn new<'b>(state: &'b mut ProcessorState<T>, handler: &'b mut H) -> Performer<'b, H, T> {
+        Performer { state, handler }
+    }
+}
+
+#[cfg(not(feature = "no_std"))]
+#[derive(Default)]
+pub struct StdSyncHandler {
+    timeout: Option<std::time::Instant>,
```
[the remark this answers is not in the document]
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/34d228cfd0cf7d9ef9fb5d5f0203323c6f3f76ab.

Cargo.toml
Andy-Python-Programmer: >f64 are included even in core though, right? 

Yeah they are available in core, but in a kernel you do not want to use floating point numbers because they are tedious to use. 

>I'm not aware of any crate offering explicit exclusion for floating point numbers.

If you would like to remove the feature, I could and just patch out functions that make use of floating points when I am using the crate.

src/ansi.rs
```diff
@@ -0,0 +1,2064 @@
...
+    #[inline]
+    fn pending_timeout(&self) -> bool {
+        self.timeout.is_some()
+    }
+}
+
+pub trait SyncHandler: Default {
+    /// Expiration time of the synchronized update.
+    fn update_timeout(&mut self, _: Option<Duration>);
+    fn pending_timeout(&self) -> bool;
```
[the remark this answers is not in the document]
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/c48fcd8986d1a6df2f8b8ba7de83bcaa8c855bea.

src/ansi.rs
```diff
@@ -0,0 +1,2064 @@
...
+    pub fn sync_timeout(&self) -> Option<std::time::Instant> {
+        self.timeout
+    }
+}
+
+#[cfg(not(feature = "no_std"))]
+impl SyncHandler for StdSyncHandler {
+    fn update_timeout(&mut self, duration: Option<Duration>) {
+        use std::time::Instant;
+        self.timeout = duration.map(|e| Instant::now() + e);
```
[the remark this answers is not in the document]
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/c48fcd8986d1a6df2f8b8ba7de83bcaa8c855bea. Took the `set_timout` and `clear_timeout` approach.

src/ansi.rs
```diff
@@ -0,0 +1,2064 @@
...
+
+    #[inline]
+    fn pending_timeout(&self) -> bool {
+        self.timeout.is_some()
+    }
+}
+
+pub trait SyncHandler: Default {
+    /// Expiration time of the synchronized update.
+    fn update_timeout(&mut self, _: Option<Duration>);
```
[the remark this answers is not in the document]
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/c48fcd8986d1a6df2f8b8ba7de83bcaa8c855bea.

src/ansi.rs
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/c48fcd8986d1a6df2f8b8ba7de83bcaa8c855bea.

Cargo.toml
chrisduerr: I'm mainly not entirely sure about the concrete motivation. If you have `f64`s anyway, it shouldn't matter if you have more of them, right?

These methods are completely optional, so you don't *have* to use them.

src/ansi.rs
```diff
@@ -0,0 +1,2082 @@
...
+    #[inline]
+    fn pending_timeout(&self) -> bool {
+        self.timeout.is_some()
+    }
+}
+
+pub trait Timeout: Default {
+    /// Sets the timeout for the next synchronized update. The `duration` parameter
+    /// specifies the duration of the timeout. Once the timeout has elapsed, the
+    /// synchronized update can be performed.
```
chrisduerr, line 461:
```suggestion
    /// Sets the timeout for the next synchronized update.
    ///
    /// The `duration` parameter specifies the duration of the timeout. Once the timeout has elapsed, the synchronized update can be performed.
```

Doc comments should be one short tagline with detailed explanation in paragraphs separated by empty lines.

src/ansi.rs
```diff
@@ -0,0 +1,2082 @@
...
+    fn set_timeout(&mut self, duration: Duration);
+    /// Clear the current timeout.
+    fn clear_timeout(&mut self);
+    /// Returns whether a timeout is currently pending.
```
chrisduerr, line 465: I think this is slightly unclear. Does this mean the timeout has run out, so it's pending the cancellation, or does it mean there is a timeout ongoing (which isn't over yet).

It's kinda confusing because `timeout` is both a verb and a noun. We want to use `timeout` as a verb here but that might need clarification in the docs. I'm happy to change the name too if you have a better idea.

src/ansi.rs
```diff
@@ -0,0 +1,2082 @@
...
+        self.timeout = None;
+    }
+
+    #[inline]
+    fn pending_timeout(&self) -> bool {
+        self.timeout.is_some()
+    }
+}
+
+pub trait Timeout: Default {
```
chrisduerr, line 458: A doc comment on the trait in general would be helpful for implementors too I think.

Just the gist so developers get the point that this is for timer implementations.

chrisduerr (changes requested):
Updating the Rust version is no problem at all, 1.56 is pretty old.

Could you add a changelog entry documenting this new feature plus stating that the MSRV has changed? Just adding it under an `Unreleased` entry should ensure we don't forget on release.

Cargo.toml
Andy-Python-Programmer: I misinterpreted the error message from the compiler, which indicated that `f64::powf` was missing in `f64`. However, it wasn't due to the target's lack of support, but rather because it wasn't accessible in `no_std`. But still, it would be required to be gated under `no_std`.

src/ansi.rs
Andy-Python-Programmer:
```suggestion
    /// Returns whether a timeout is currently active and has not yet expired.
```

src/ansi.rs
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/b91f035c6b731e5162d9099808e70fce1a99167c

src/ansi.rs
Andy-Python-Programmer: Fixed in https://github.com/alacritty/vte/pull/85/commits/b91f035c6b731e5162d9099808e70fce1a99167c.

Cargo.toml
Andy-Python-Programmer: This should be resolved in https://github.com/alacritty/vte/pull/85/commits/18cc86b9e80a9c3a0fb2d72ce86003e04941d1b2 (now only available when `std` is available).

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+- Minimum rust version has been bumped to 1.62.1
+- Support for ANSI terminal stream parsing under the `ansi` feature.
+- Addition of the `serde` feature which additionally derives `Serialize` and `Deserialize` 
+  for the types provided in the `ansi` module.
```
chrisduerr, line 9:
```suggestion
- Addition of the `serde` feature which derives `Serialize` and `Deserialize`
  for the types provided in the `ansi` module.
```

Double "addition" is a bit repetitive here.

src/ansi.rs
```diff
@@ -0,0 +1,2089 @@
...
+    }
+
+    #[inline]
+    fn pending_timeout(&self) -> bool {
+        self.timeout.is_some()
+    }
+}
+
+/// Provides an interface for common timer functionality (i.e.
+/// creating timeouts and checking if it has been expired).
```
chrisduerr, line 459:
```suggestion
/// Interface for creating timeouts and checking their expiry.
```

The whole parenthesis situation probably doesn't generate the best docs and reads a bit awkward. I think this one simple line should say the same thing but look a bit cleaner.

chrisduerr (changes requested):

chrisduerr (approved):
Thanks for your work.

Andy-Python-Programmer:
:^) np

PR #91, opened 2023-07-15
Allow no-std users to control the size of the osc_raw buffer.

### LINKED ISSUE #90: Limiting size of osc_raw in no-std mode
I am writing an OS and I was in need of an ANSI parser. I started to write my own, and then was pointed at your excellent crate. Thank you!

I note that `const MAX_OSC_RAW: usize = 1024` and that is a little large for me. Would you accept a PR to make it a const-generic? Something like:

```rust
/// Parser for raw _VTE_ protocol which delegates actions to a [`Perform`]
///
/// [`Perform`]: trait.Perform.html
#[derive(Default)]
pub struct Parser<const OSC_RAW_SIZE: usize = MAX_OSC_RAW> {
    state: State,
    intermediates: [u8; MAX_INTERMEDIATES],
    intermediate_idx: usize,
    params: Params,
    param: u16,
    #[cfg(feature = "no_std")]
    osc_raw: ArrayVec<u8, OSC_RAW_SIZE>,
    #[cfg(not(feature = "no_std"))]
    osc_raw: Vec<u8>,
    osc_params: [(usize, usize); MAX_OSC_PARAMS],
    osc_num_params: usize,
    ignoring: bool,
    utf8_parser: utf8::Parser,
}
```

The const-generic parameter will be there even in no-std mode, but as it has a default, I don't think it'll matter?

thejpster:
It was set to 1024 bytes, which is a lot if you don't care about parsing Operating System Command sequences. Now you can set it to a custom size.

Closes #90

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,7 +52,6 @@ use definitions::{unpack, Action, State};
 
 const MAX_INTERMEDIATES: usize = 2;
 const MAX_OSC_PARAMS: usize = 16;
-#[cfg(any(feature = "no_std", test))]
 const MAX_OSC_RAW: usize = 1024;
 
 struct VtUtf8Receiver<'a, P: Perform>(&'a mut P, &'a mut State);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -72,15 +71,19 @@ impl<'a, P: Perform> utf8::Receiver for VtUtf8Receiver<'a, P> {
 /// Parser for raw _VTE_ protocol which delegates actions to a [`Perform`]
 ///
 /// [`Perform`]: trait.Perform.html
+///
+/// Generic over the value for the size of the raw Operating System Command
+/// buffer, but that only has an effect in when the `no_std` feature is enabled
+/// and can be ignored otherwise.
```
chrisduerr, line 77:
```suggestion
/// Generic over the value for the size of the raw Operating System Command
/// buffer. Only used when the `no_std` feature is enabled.
```

src/lib.rs
```diff
@@ -923,6 +940,28 @@ mod tests {
...
+        match &dispatcher.dispatched[0] {
+            Sequence::Csi(params, intermediates, ignore, _) => {
+                assert_eq!(intermediates, &[b'?']);
+                assert_eq!(params, &[[1049]]);
+                assert!(!ignore);
+            },
+            _ => panic!("expected csi sequence"),
+        }
+    }
 }
```
chrisduerr, line 965: It would probably make sense to test this with OSC escapes. Ideally even to check that it breaks when the limit is exceeded.

chrisduerr (changes requested):

src/lib.rs
thejpster: Added some tests. Apparently the behaviour is to silently truncate when the limit is exceeded - unsure if that's intended or not.

chrisduerr: It is. The limit is supposed to be the upper limit of what you'll ever use. If that isn't big enough for your purposes, that's on you for setting a limit that is too small.

chrisduerr (approved):
LGTM, thanks for the PR.

PR #92, opened 2023-07-30
Add support for OSC 22

kchibisov:


src/ansi.rs
```diff
@@ -668,6 +668,9 @@ pub trait Handler {
 
     /// Set hyperlink.
     fn set_hyperlink(&mut self, _: Option<Hyperlink>) {}
+
+    /// Set mouse cursor.
+    fn set_mouse_cursor(&mut self, _: String) {}
```
chrisduerr, line 673: How are consumers supposed to know what this string means? There isn't even any documentation for it.

At the very least we should document the possible fields, or provide an enum with all possible values.

chrisduerr (changes requested):

src/ansi.rs
kchibisov: I've updated the docs.

We could enforce the naming with the `cursor-icon` if it matters.

chrisduerr (changes requested):
Nothing was changed.

src/ansi.rs
kchibisov: Hm, github just bugged, because the git push said that everything up to date, but the commit wasn't matching the one on github, ammend + force push resolved though.

chrisduerr: I had the same thing happen to me the other day, it's really strange.

src/ansi.rs
```diff
@@ -668,6 +668,16 @@ pub trait Handler {
 
     /// Set hyperlink.
     fn set_hyperlink(&mut self, _: Option<Hyperlink>) {}
+
+    /// Set mouse cursor.
+    ///
+    /// The requested cursor shapes usually have a name from CSS specificiation
+    /// <https://www.w3.org/TR/css-ui-3/#cursor>.
```
chrisduerr, line 675: I think `usually` phrasing is a bit odd here. Of course a client can pick whatever, but it's a bit too uncertain.

I think it would be better to phrase it more like this:

```suggestion
    /// Clients can specify any string, but legal values are defined in the CSS specificiation:
    /// <https://www.w3.org/TR/css-ui-3/#cursor>.
```

But there's really not that many values available. Why wouldn't we just define an enum? That would make everything **so** much easier for consumers.

src/ansi.rs
```diff
@@ -668,6 +668,16 @@ pub trait Handler {
...
+    ///
+    /// Some legacy clients could send a legacy X11 names, but it's encouradges
+    /// to use CSS values and fallback to default cursor or ignore requets when
+    /// the name is not the one from the said specification.
```
chrisduerr, line 679:
```suggestion
    /// Some legacy clients could send a legacy X11 name, but it's encouraged
    /// to use CSS values and fall back to the default cursor or ignore requests when
    /// the name is not one from the said specification.
```

I'd also say that "fall back to default or ignore requests" is not precise enough. There should be **one** correct behavior here, not two. This must be clearly defined.

chrisduerr (changes requested):

src/ansi.rs
```diff
@@ -1267,6 +1271,16 @@ where
                 unhandled(params);
             },
 
+            // Set mouse cursor shape.
+            b"22" if params.len() == 2 => {
+                let shape = String::from_utf8_lossy(params[1]).to_string();
+                if let Ok(cursor_icon) = CursorIcon::from_str(&shape) {
+                    self.handler.set_mouse_cursor_icon(cursor_icon);
+                } else {
+                    debug!("[osc 22] unrecognized cursor icon shape {shape}");
```
chrisduerr, line 1280:
```suggestion
                    debug!("[osc 22] unrecognized cursor icon shape: {shape:?}");
```

src/ansi.rs
```diff
@@ -1267,6 +1271,16 @@ where
                 unhandled(params);
             },
 
+            // Set mouse cursor shape.
+            b"22" if params.len() == 2 => {
+                let shape = String::from_utf8_lossy(params[1]).to_string();
+                if let Ok(cursor_icon) = CursorIcon::from_str(&shape) {
```
chrisduerr, line 1277: Should match rather than if let else.

chrisduerr (changes requested):
Looks good other than these simple style changes.

chrisduerr (approved):

PR #93, opened 2023-07-31
Add support for kitty keyboard protocol (CSI u)

kchibisov:
This just adds handling of the mentioned CSI escapes in the spec and don't provide any sequence transformation for the terminals

Links: https://sw.kovidgoyal.net/kitty/keyboard-protocol


kchibisov:
>Hm, also needs CSI = flags ; mode u

Done.

chrisduerr (changes requested):
Seems like I forgot to submit this?

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+- Add support for kitty protocol (`CSI u`)
```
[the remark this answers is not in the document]
kchibisov: There's different `CSI u` from fixterms.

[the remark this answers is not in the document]
kchibisov: That's also `CSI u`, but completely different for example http://www.leonerd.org.uk/hacks/fixterms/

And kitty is the one defining the query/stack stuff.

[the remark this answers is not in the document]
chrisduerr: Yeah, `libvterm` is written by leonerd, that's what I was talking about.

I'd just remove mention of the `CSI u` since it might confuse people.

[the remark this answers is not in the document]
kchibisov: That's fair.

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+
+    /// Report current keyboard mode.
+    ///
+    /// ## Caveat
+    ///
+    /// This request must be done successively with the [`identify_terminal`] request, given
+    /// clients rely on the later to identify whether the terminal suppors kitty's keyboard
+    /// protocol.
+    ///
+    /// [`identify_terminal`]: Self::identify_terminal
```
chrisduerr, line 685: This technically isn't actually true. The whole `identify_terminal` thing is just a hack for applications to figure out if an escape sequence is not supported, or if there just hasn't been a reply yet. Basically just a "sync point" and other escapes could work too.

This is very much a general thing, even though this protocol specifically incorporates it, so I don't think this section is necessary. People curious about the exact correct behavior are probably better off reading the protocol itself anyway.

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+    fn report_keyboard_mode(&mut self) {}
+
+    /// Push keyboard mode into the keyboard mode stack.
+    fn push_keyboard_mode(&mut self, _mode: KeyboardMode) {}
+
+    /// Pop the given amount of keyboard modes from the
+    /// keyboard mode stack.
+    fn pop_num_keyboard_modes(&mut self, _to_pop: u16) {}
```
chrisduerr, line 693:
```suggestion
    /// Pop the given amount of keyboard modes from the
    /// keyboard mode stack.
    fn pop_keyboard_modes(&mut self, _to_pop: u16) {}
```

I think the separation between "mode" and "modes" should be enough to signify multiple are being popped. The name is a bit awkward and long otherwise and I don't think there's ever going to be a chance of confusion anyway.

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+    /// Set the [`keyboard mode`] using the given [`behavior`].
+    ///
+    /// [`keyboard mode`]: crate::ansi::KeyboardMode
+    /// [`behavior`]: crate::ansi::KeyboardModeApplyBehavior
+    fn set_keyboard_mode(&mut self, _mode: KeyboardMode, _behavior: KeyboardModeApplyBehavior) {}
+}
+
+bitflags! {
+    /// The keyboard mode representing [`kitty keyboard protocol'].
+    ///
```
chrisduerr, line 704: These /// are different?! Is GitHub's code view completely broken now?

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+    /// The details on how to build the escape sequences could be found in the
+    /// mentioned specification.
+    ///
+    /// [`kitty keyboard protocol']: https://sw.kovidgoyal.net/kitty/keyboard-protocol
+    #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    pub struct KeyboardMode : u8 {
```
chrisduerr, line 710: I think having this as singular is confusing since it's multiple modes stored at the same time.

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+
+    /// Set the [`keyboard mode`] using the given [`behavior`].
+    ///
+    /// [`keyboard mode`]: crate::ansi::KeyboardMode
+    /// [`behavior`]: crate::ansi::KeyboardModeApplyBehavior
+    fn set_keyboard_mode(&mut self, _mode: KeyboardMode, _behavior: KeyboardModeApplyBehavior) {}
+}
+
+bitflags! {
+    /// The keyboard mode representing [`kitty keyboard protocol'].
```
chrisduerr, line 703:
```suggestion
    /// A set of  [`kitty keyboard protocol'] modes.
```

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+    ///
+    /// The details on how to build the escape sequences could be found in the
+    /// mentioned specification.
+    ///
```
chrisduerr, line 707:
```suggestion
```

I don't think this is necessary, especially because the mode isn't really directly related to the escape sequence.

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+    /// [`kitty keyboard protocol']: https://sw.kovidgoyal.net/kitty/keyboard-protocol
+    #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    pub struct KeyboardMode : u8 {
+        /// No particular mode is being set.
+        const NO_MODE                 = 0b0000_0000;
+        /// Report `Esc`, `alt` + `key`, `ctrl` + `key`, `ctrl` + `alt` + `key`, `shift`
+        /// + `alt` + `key` keys using `CSI u` sequence instead of legacy ones.
+        const DISAMBIGUATE_ESC_CODES  = 0b0000_0001;
+        /// Report key presses, release, repeat along side the escape. If the
```
chrisduerr, line 716:
```suggestion
        /// Report key presses, release, and repetition alongside the escape. If the
```

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+        /// key event results in text the type is not support for it, unless
+        /// [`REPORT_ALL_KEYS_AS_ESC`] is enabled.
```
chrisduerr, line 718: What does this mean? "the type is not support for it" specifically.

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+        const REPORT_EVENT_TYPES      = 0b0000_0010;
+        /// Report alternative value in pair with original value.
```
chrisduerr, line 720:
```suggestion
        /// Additionally report shifted key and base layout key.
```

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+    /// The details on how to build the escape sequences could be found in the
+    /// mentioned specification.
+    ///
+    /// [`kitty keyboard protocol']: https://sw.kovidgoyal.net/kitty/keyboard-protocol
+    #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    pub struct KeyboardMode : u8 {
+        /// No particular mode is being set.
+        const NO_MODE                 = 0b0000_0000;
+        /// Report `Esc`, `alt` + `key`, `ctrl` + `key`, `ctrl` + `alt` + `key`, `shift`
+        /// + `alt` + `key` keys using `CSI u` sequence instead of legacy ones.
```
chrisduerr, line 714:
```suggestion
        /// Report `Esc`, `alt` + `key`, `ctrl` + `key`, `ctrl` + `alt` + `key`, `shift`
        /// + `alt` + `key` keys using `CSI u` sequence instead of raw ones.
```

Not using this protocol isn't really "legacy". It is and always will be one of the easiest and most convenient ways to handle text input for terminal applications.

I've picked "raw" here because it was the first thing that came to my mind, but I'm open for alternative suggestions.

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+bitflags! {
+    /// The keyboard mode representing [`kitty keyboard protocol'].
+    ///
+    /// The details on how to build the escape sequences could be found in the
+    /// mentioned specification.
+    ///
+    /// [`kitty keyboard protocol']: https://sw.kovidgoyal.net/kitty/keyboard-protocol
+    #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    pub struct KeyboardMode : u8 {
+        /// No particular mode is being set.
```
chrisduerr, line 711:
```suggestion
        /// No keyboard protocol mode is set.
```

src/ansi.rs
```diff
@@ -672,6 +673,70 @@ pub trait Handler {
...
+        const REPORT_ALL_KEYS_AS_ESC  = 0b0000_1000;
+        /// Report the text generated by the key event.
+        const REPORT_ASSOCIATED_TEXT  = 0b0001_0000;
+    }
+}
+
+/// Describes how the new [`KeyboardMode`] should be applied.
+#[repr(u8)]
+#[derive(Default, Clone, Copy, PartialEq, Eq)]
+pub enum KeyboardModeApplyBehavior {
```
chrisduerr, line 732: Generally I'd say "ApplicationBehavior" is more correct here, but I think it's better to leave it as "Apply" since application has multiple meanings.

chrisduerr (changes requested):

src/ansi.rs
kchibisov: seems like it.

src/ansi.rs
kchibisov: That the event type information won't be attached to e.g. `e`.

chrisduerr (approved):

PR #95, opened 2023-09-17
Add support for xterm's modifyOtherKeys CSI

kchibisov:


src/ansi.rs
```diff
@@ -1525,6 +1595,15 @@ where
                     }
                 }
             },
+            ('m', [b'>']) => {
+                let mode = match (next_param_or(1) == 4).then(|| next_param_or(0)) {
```
chrisduerr, line 1579: Using `next_param_or(1) == 4` seems very odd to me.

Probably better to just do the following:

```
if let Some(&[4]) = params_iter.next() {
    …
}
```

Or some variation of this (there's probably better options that don't indent so heavily).

src/ansi.rs
```diff
@@ -1525,6 +1595,15 @@ where
                     }
                 }
             },
+            ('m', [b'>']) => {
```
chrisduerr, line 1578: Might want to also support `CSI ? Pp m`?

src/ansi.rs
```diff
@@ -689,6 +689,9 @@ pub trait Handler {
     /// [`keyboard mode`]: crate::ansi::KeyboardModes
     /// [`behavior`]: crate::ansi::KeyboardModesApplyBehavior
     fn set_keyboard_mode(&mut self, _mode: KeyboardModes, _behavior: KeyboardModesApplyBehavior) {}
+
+    /// Set xterm's [`ModifyOtherKeys`] option.
```
chrisduerr, line 693:
```suggestion
    /// Set XTerm's [`ModifyOtherKeys`] option.
```

src/ansi.rs
```diff
@@ -715,6 +718,73 @@ bitflags! {
     }
 }
 
+/// Construct escape sequences for regular keys, like `1`, when
+/// modified by `Shift-`, `Control-`, `Alt-` or `Meta-`modifiers.
+/// This doesn't apply to the named keys on the keyboard, like `F1`.
+/// The escape sequence also doesn't apply to defined unicode input, like
+/// cyrillic chars.
```
chrisduerr, line 725: The "defined" here is confusing. What does "defined unicode input" mean?

Is the following correct?

```
/// This only applies to keys corresponding to ascii characters.
```

If it is, this would be simpler and it would also remove the need for the `F1` comment.

src/ansi.rs
```diff
@@ -715,6 +718,73 @@ bitflags! {
...
+///
+/// For the details on how to implement the mode handling correctly
+/// consult the xterm's [`implementation`] and the output of the
+/// xterm's provided perl [`script`].
```
chrisduerr:
```suggestion
/// For the details on how to implement the mode handling correctly,
/// consult [`XTerm's implementation`] and the output of XTerm's provided [`perl script`].
```

src/ansi.rs
```diff
@@ -715,6 +718,73 @@ bitflags! {
...
+/// [`script`]: https://github.com/ThomasDickey/xterm-snapshots/blob/master/vttests/modify-keys.pl
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ModifyOtherKeys {
+    /// Reset the state.
+    Reset,
+    /// Enables this feature for keys except for those with well-
+    /// known behavior, e.g., Tab, Backarrow and some special
+    /// control character cases which are built into the X11
+    /// library, e.g., Control-Space to make a NUL, or Control-3
+    /// to make an Escape character.
```
chrisduerr, line 743:
```suggestion
    /// Enables this feature except for keys with well-
    /// known behavior, e.g., Tab, Backspace and some special
    /// control character cases which are built into the X11
    /// library (e.g. Control-Space to make a NUL, or Control-3
    /// to make an Escape character).
```

src/ansi.rs
```diff
@@ -715,6 +718,73 @@ bitflags! {
...
+    ///
+    /// The following cases could be outlined based on the xterm
+    /// implementation when the escape sequences shouldn't be
+    /// created:
```
chrisduerr, line 747: The wording here is a bit strange. If the following is correct, it would be better:

```
Escape sequences shouldn't be emitted under the following circumstances:
```

src/ansi.rs
```diff
@@ -715,6 +718,73 @@ bitflags! {
...
+    ///
+    /// - When the key is in range of `[0x40;0x7f]` and the modifier is either Control or Shift;
```
chrisduerr, line 749: Using hex codes here kinda makes things more complicated than they have to be.

src/ansi.rs
```diff
@@ -715,6 +718,73 @@ bitflags! {
...
+    /// - when the key combination is a known control combination alias, and when only Shift or
+    ///   Control is pressed:
```
chrisduerr: Inconsistent campitalization of `when`, I don't care either way but should be consistent.

```suggestion
    /// - When the key combination is a known control combination alias, and only Shift or
    ///   Control is pressed:
```

src/ansi.rs
```diff
@@ -715,6 +718,54 @@ bitflags! {
...
+/// cyrillic chars.
+///
+/// For the details on how to implement the mode handling correctly
+/// consult the xterm's [`implementation`] and the output of the
+/// xterm's provided perl [`script`]. An example of the output
+/// is present [`here`].
+///
+/// [`implementation`]: https://invisible-island.net/xterm/modified-keys.html
+/// [`script`]: https://github.com/ThomasDickey/xterm-snapshots/blob/master/vttests/modify-keys.pl
+/// [`here`]: https://github.com/alacritty/vte/blob/master/doc/modifyOtherKeys-example.txt
```
kchibisov, line 734: I can't do relative link, but that should be good enough.

src/ansi.rs
chrisduerr: You removed the tables, but this line is still referencing something that should come after it using the colon.

src/ansi.rs
```diff
@@ -715,6 +718,54 @@ bitflags! {
...
+    ///
+    /// The following cases could be outlined based on the xterm
+    /// implementation when the escape sequences shouldn't be
+    /// created:
+    ///
+    /// - When the key is in range of `[0x40;0x7f]` and the modifier is either Control or Shift;
+    /// - when the key combination is a known control combination alias, and when only Shift or
+    ///   Control is pressed:
+    ///
+    /// For more details, consult the [`example`] of the suggested translation.
```
chrisduerr, line 752:
```suggestion
    /// For more details, consult the [`example`] for the suggested translation.
```

src/ansi.rs
```diff
@@ -715,6 +718,53 @@ bitflags! {
...
+    /// The following cases could be outlined based on the xterm
+    /// implementation when the escape sequences shouldn't be
+    /// created:
+    ///
+    /// - When the key is in range of `[0x40;0x7f]` and the modifier is either Control or Shift;
+    /// - when the key combination is a known control combination alias.
+    ///
+    /// For more details, consult the [`example`] of the suggested translation.
+    ///
+    /// [`example`]: https://github.com/alacritty/vte/blob/master/doc/modifyOtherKeys-example.txt
```
chrisduerr, line 754: I don't think there's any way to link to the filesystem here, is there? This link seems kinda easy to break.

src/ansi.rs
```diff
@@ -715,6 +718,53 @@ bitflags! {
...
+    EnableExceptWellDefined,
+    /// Enables this feature for keys including the exceptions
+    /// listed.  Xterm ignores the special cases built into the
+    /// X11 library.  Any shifted (modified) ordinary key sends an
+    /// escape sequence.  The Alt- and Meta- modifiers cause xterm
+    /// to send escape sequences.
```
chrisduerr, line 760:
```suggestion
    /// Enables this feature for all keys including the exceptions
    /// of [`Self::EnableExceptWellDefined`]. XTerm still ignores the special cases built into the
    /// X11 library. Any shifted (modified) ordinary key send an
    /// escape sequence. The Alt- and Meta- modifiers cause XTerm
    /// to send escape sequences.
```

chrisduerr (changes requested):

src/ansi.rs
kchibisov: I used the xterm docs here, if it matters.

src/ansi.rs
kchibisov: there's no such thing anymore.

src/ansi.rs
kchibisov: yeah, there's no way to do so. We could write an alternative markdown document and include it, but I think it's for the entire enum and it'll be very verbose.

src/ansi.rs
chrisduerr: Not the worst I suppose, I still prefer my suggestions. :P

src/ansi.rs
kchibisov: I'm not sure you can do any better than what I use, simply because you'd need mulitple `unhandled!` cases.

kchibisov: Nah, I don't like that, it's just bad, because the closure holds the borrow and you'd have to clone/map. We have an example bellow which looks like:

```rust
                let bottom =
                    params_iter.next().map(|param| param[0] as usize).filter(|&param| param != 0);

```

I'd basically need the same just to check that it's `4`.

chrisduerr: Ah `next_param_or` captures `params_iter` quite unfortunately.

src/ansi.rs
```diff
@@ -715,6 +723,44 @@ bitflags! {
     }
 }
 
+/// This only applies to keys corresponding to ascii characters.
+///
+/// For the details on how to implement the mode handling correctly, consult [`XTerm's
+/// implementation`] and the output of XTerm's provided [`perl script`], which output
+/// example could be found [`here`].
```
chrisduerr, line 730:
```suggestion
/// For the details on how to implement the mode handling correctly, consult [`XTerm's
/// implementation`] and the output of XTerm's provided [`perl script`], the output of which
/// can be found [`here`].
```

Alternatively whose also works. "an example of its output can be found here" is another option.

src/ansi.rs
```diff
@@ -1525,6 +1571,22 @@ where
...
+                    _ => return unhandled!(),
+                };
+                handler.set_modify_other_keys(mode);
+            },
+            ('m', [b'?']) => {
+                if params_iter.next() == Some(&[4]) {
+                    handler.report_modify_other_keys();
+                    return;
+                }
+                unhandled!();
```
chrisduerr, line 1588:
```suggestion
                if params_iter.next() == Some(&[4]) {
                    handler.report_modify_other_keys();
                } else {
                    unhandled!();
                }
```

Ton of control flow complexity in this file, probably better to avoid return.

src/ansi.rs
```diff
@@ -689,6 +689,14 @@ pub trait Handler {
...
     /// [`behavior`]: crate::ansi::KeyboardModesApplyBehavior
     fn set_keyboard_mode(&mut self, _mode: KeyboardModes, _behavior: KeyboardModesApplyBehavior) {}
+
+    /// Set XTerm's [`ModifyOtherKeys`] option.
+    fn set_modify_other_keys(&mut self, _mode: ModifyOtherKeys) {}
+
+    /// Report XTerm's [`ModifyOtherKeys`] option.
+    ///
+    /// The output is of form `CSI > 4 : mode m`.
+    fn report_modify_other_keys(&mut self) {}
```
chrisduerr, line 699:
```suggestion
    /// Report XTerm's [`ModifyOtherKeys`] state.
    ///
    /// The output is of form `CSI > 4 ; mode m`.
    fn report_modify_other_keys(&mut self) {}
```

chrisduerr (changes requested):

kchibisov:
I'd need to understand what real world apps want, maybe they want CSI u in responce and not 27 variant, since in #neovim channel I was told to never use 27 variant.

kchibisov:
I've updated docs to connect all the relevant information on how to implement, also I've regenerated the XTerm's script output to be with `CSI u` syntax.

src/ansi.rs
```diff
@@ -678,6 +686,50 @@ bitflags! {
     }
 }
 
+/// This only applies to keys corresponding to ascii characters.
```
chrisduerr, line 689: Usually the first line is a headline on what the struct is actually about.

So maybe make this the first paragraph and add a first line like this:

```
/// XTMODKEYS modifyOtherKeys state.
```

src/ansi.rs
```diff
@@ -678,6 +686,50 @@ bitflags! {
...
+///
+/// For the details on how to implement the mode handling correctly, consult [`XTerm's
+/// implementation`] and the output of XTerm's provided [`perl script`], the output of
+/// which for `CSI u` variant can be found [`here`]. Some libraries and implementations
```
chrisduerr, line 693:
```suggestion
/// implementation`] and the [`output`] of XTerm's provided [`perl script`]. Some libraries and implementations
```

Shorter!

src/ansi.rs
```diff
@@ -678,6 +686,50 @@ bitflags! {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ModifyOtherKeys {
+    /// Reset the state.
+    Reset,
+    /// Enables this feature except for keys with well-known behavior, e.g., Tab, Backspace and
+    /// some special control character cases which are built into the X11 library (e.g.,
+    /// Control-Space to make a NUL, or Control-3 to make an Escape character).
+    ///
+    /// Escape sequences shouldn't be emitted under the following circumstances:
+    ///
```
chrisduerr, line 714:
```suggestion
```

Just for brevity, don't think it's necessary. Not a big deal either way.

src/ansi.rs
```diff
@@ -678,6 +686,50 @@ bitflags! {
...
+    /// - When the key is in range of `[64;127]` and the modifier is either Control or Shift;
+    /// - when the key combination is a known control combination alias.
```
chrisduerr, line 716:
```suggestion
    /// - When the key is in range of `[64;127]` and the modifier is either Control or Shift
    /// - When the key combination is a known control combination alias
```

It's a list, shouldn't be written like a sentence split apart with bullet points.

chrisduerr (changes requested):
Ready to go if you just apply my suggestions.

src/ansi.rs
kchibisov: depends, it's sort of common to make it a `;` separated in papers and have a e.g. dot in the last bullet point, but I don't really mind this thing in docs.

src/ansi.rs
```diff
@@ -678,6 +686,48 @@ bitflags! {
...
+pub enum ModifyOtherKeys {
+    /// Reset the state.
+    Reset,
+    /// Enables this feature except for keys with well-known behavior, e.g., Tab, Backspace and
+    /// some special control character cases which are built into the X11 library (e.g.,
+    /// Control-Space to make a NUL, or Control-3 to make an Escape character).
+    ///
+    /// Escape sequences shouldn't be emitted under the following circumstances:
+    /// - When the key is in range of `[64;127]` and the modifier is either Control or Shift
+    /// - When the key combination is a known control combination alias.
```
chrisduerr, line 714:
```suggestion
    /// - When the key combination is a known control combination alias
```

src/ansi.rs
```diff
@@ -678,6 +686,48 @@ bitflags! {
     }
 }
 
+/// XTMODKEYS modifyOtherKeys state. This only applies to keys corresponding to ascii characters.
```
chrisduerr, line 689:
```suggestion
/// XTMODKEYS modifyOtherKeys state.
///
/// This only applies to keys corresponding to ascii characters.
```

chrisduerr (changes requested):

chrisduerr (approved):

PR #96, opened 2023-09-18
Switch sync escape from DCS to CSI

chrisduerr:
See alacritty/alacritty#6845.

### DIFF
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -39,14 +39,14 @@ const SYNC_UPDATE_TIMEOUT: Duration = Duration::from_millis(150);
 /// Maximum number of bytes read in one synchronized update (2MiB).
 const SYNC_BUFFER_SIZE: usize = 0x20_0000;
 
-/// Number of bytes in the synchronized update DCS sequence before the passthrough parameters.
-const SYNC_ESCAPE_START_LEN: usize = 5;
+/// Number of bytes in the BSU/ESU CSI sequences.
+const SYNC_ESCAPE_LEN: usize = 8;
 
-/// Start of the DCS sequence for beginning synchronized updates.
-const SYNC_START_ESCAPE_START: [u8; SYNC_ESCAPE_START_LEN] = [b'\x1b', b'P', b'=', b'1', b's'];
+/// BSU CSI sequence for beginning or extending synchronized updates.
+const BSU_CSI: [u8; SYNC_ESCAPE_LEN] = *b"\x1b[?2026h";
 
-/// Start of the DCS sequence for terminating synchronized updates.
-const SYNC_END_ESCAPE_START: [u8; SYNC_ESCAPE_START_LEN] = [b'\x1b', b'P', b'=', b'2', b's'];
+/// ESU CSI sequence for terminating synchronized updates.
+const ESU_CSI: [u8; SYNC_ESCAPE_LEN] = *b"\x1b[?2026l";
 
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub struct Hyperlink {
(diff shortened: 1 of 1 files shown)


src/ansi.rs
```diff
@@ -363,46 +343,29 @@ impl<T: Timeout> Processor<T> {
...
+    {
         // Get the last few bytes for comparison.
         let len = self.state.sync_state.buffer.len();
-        let offset = len.saturating_sub(SYNC_ESCAPE_START_LEN);
+        let offset = len.saturating_sub(SYNC_ESCAPE_LEN);
         let end = &self.state.sync_state.buffer[offset..];
 
+        // NOTE: It is technically legal to specify multiple private modes in the same
+        // escape, but we only allow EXACTLY `\e[?2026h`/`\e[?2026l` to keep the parser
+        // reasonable.
```
chrisduerr, line 362: I've tested this with Kitty (and read their impl) and they do the same. Seems like somewhat of an oversight of the protocol specification.

src/ansi.rs
```diff
@@ -812,6 +775,8 @@ pub enum Mode {
     SwapScreenAndSetRestoreCursor = 1049,
     /// ?2004
     BracketedPaste = 2004,
+    /// Synchronized update start/end.
+    SyncUpdate = 2026,
```
chrisduerr, line 779: This forces Alacritty to handle this mode in our `set_mode`/`unset_mode` match statement. But I didn't want to handle this mode completely separately, so it's probably fine?

kchibisov: The thing is that it's confusing for the consumers, because this mode doesn't tell a thing.

Though, we can easily check for it separately. It just strange that we have a mode that can never be populated to downstream and that they could think that they need to freeze the grid or something like that.

kchibisov (changes requested):

src/ansi.rs
```diff
@@ -39,14 +39,14 @@ const SYNC_UPDATE_TIMEOUT: Duration = Duration::from_millis(150);
 /// Maximum number of bytes read in one synchronized update (2MiB).
 const SYNC_BUFFER_SIZE: usize = 0x20_0000;
 
-/// Number of bytes in the synchronized update DCS sequence before the passthrough parameters.
-const SYNC_ESCAPE_START_LEN: usize = 5;
+/// Number of bytes in the BSU/ESU CSI sequences.
+const SYNC_ESCAPE_LEN: usize = 8;
```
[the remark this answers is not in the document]
chrisduerr: I'm not sure what you mean? This way we have a constant to use in the size of the byte arrays, otherwise we'd have no way to guarantee both arrays stay the same length, which is important.

src/ansi.rs
chrisduerr: Yeah that's fair.

kchibisov (approved):
I don't care that much about the constant argument, I just don't know why we can't use other constant when we initialize the `_LEN` constant like `const SYNC_ESCAPE_LEN: usize = BSU_CSI.len();`, so it'll be simpler (we could also ensure that the sizes of start and end are the same in tests for example...

src/ansi.rs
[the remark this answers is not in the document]
chrisduerr: There's two arrays though. I can't define a single constant based on the length of two arrays. But I can define two arrays using a single length constant.

They **must** have the same length, so we should use a constant to synchronize them.

chrisduerr:
> we could also ensure that the sizes of start and end are the same in tests for example

Why add a test, when we get the same guarantees like this without one? It's just pointless additional code.

PR #104, opened 2023-11-14
Add support for DECRQM/DECRPM

### LINKED ISSUE #100: Add support for DECRQM
It's used to report whether mode is supported/set, etc.

This is used by sync updates as a way to detect the feature support. See https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036?permalink_comment_id=3946967#feature-detection

The mode could still be used without it, but it probably worth adding support for this escape in general.

kchibisov:
Fixes #100.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 - Reexport `cursor_icon` crate in `ansi`
 - Split-out private modes from `Mode` into `PrivateMode`
 - Add `unset_private_mode` and `set_private_mode`
+- Add `report_mode` and `report_private_mode` to handle DECRPM/DECRQM
 
 ## 0.12.0
 
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -585,11 +585,17 @@ pub trait Handler {
     /// Unset mode.
     fn unset_mode(&mut self, _mode: Mode) {}
 
+    /// DECRPM - report mode.
+    fn report_mode(&mut self, _mode: Mode) {}
+
     /// Set private mode.
     fn set_private_mode(&mut self, _mode: PrivateMode) {}
 
     /// Unset private mode.
     fn unset_private_mode(&mut self, _mode: PrivateMode) {}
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+- Implement `Clone + Copy` for `ansi::Mode
+- Add `Handler::write\_payload` to forward data from parser back to PTY
+- Add `Handler::report\_mode` handling DECRQM/DECRPM
```
chrisduerr, line 8: This is not what the function is called.

src/ansi.rs
```diff
@@ -660,6 +667,9 @@ pub trait Handler {
     ///
     /// The output is of form `CSI > 4 ; mode m`.
     fn report_modify_other_keys(&mut self) {}
+
+    /// The VTE handler requested to write payload to the terminal.
+    fn pty_write(&mut self, _payload: String) {}
```
chrisduerr, line 672: VTE writing arbitrary stuff shouldn't be a thing. There's a reason why this is a trait-based interface, we should have proper functions here.

src/ansi.rs
```diff
@@ -577,6 +581,9 @@ pub trait Handler {
     /// Set mode.
     fn set_mode(&mut self, _mode: Mode) {}
 
+    /// DECRPM - Report terminal mode.
+    fn report_mode(&mut self, _reporter: ModeReporter, _mode: Mode) {}
```
chrisduerr, line 585: Why is this passing two separate modes?

chrisduerr (changes requested):

kchibisov:
I guess this is fine? Or do we want a similar `Reporter` thing I had to help building downstream an escape sequence?

chrisduerr (approved):

kchibisov:
Added `Clone + Copy` on private mode, because it was on public, but not on private one.

PR #105, opened 2023-11-15
Split `Mode` into `Public` and `Private`

kchibisov:
The modes could overlap and there's also no way to actually forward information about unhandled modes downstream, thus split the modes into 2 separate structures and pass unhandled modes.

### DIFF
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -579,16 +579,16 @@ pub trait Handler {
     /// Set a terminal attribute.
     fn terminal_attribute(&mut self, _attr: Attr) {}
 
-    /// Set private mode.
-    fn set_public_mode(&mut self, _mode: PublicMode) {}
+    /// Set mode.
+    fn set_mode(&mut self, _mode: Mode) {}
 
-    /// Unset public mode.
-    fn unset_public_mode(&mut self, _mode: PublicMode) {}
+    /// Unset mode.
+    fn unset_mode(&mut self, _mode: Mode) {}
 
     /// Set private mode.
     fn set_private_mode(&mut self, _mode: PrivateMode) {}
 
-    /// Unset mode.
+    /// Unset private mode.
     fn unset_private_mode(&mut self, _mode: PrivateMode) {}
 
     /// DECSTBM - Set the terminal scrolling region.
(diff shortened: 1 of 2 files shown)


src/ansi.rs
```diff
@@ -574,11 +576,17 @@ pub trait Handler {
     /// Set a terminal attribute.
     fn terminal_attribute(&mut self, _attr: Attr) {}
 
-    /// Set mode.
-    fn set_mode(&mut self, _mode: Mode) {}
+    /// Set private mode.
+    fn set_public_mode(&mut self, _mode: PublicMode) {}
```
chrisduerr, line 580: Make up your mind! Is it public or private? :smile:

src/ansi.rs
```diff
@@ -770,10 +778,104 @@ pub enum CursorShape {
     Hidden,
 }
 
-/// Terminal modes.
+/// Wrapper for the public mode.
```
chrisduerr, line 781: This should reference what type of mode this is. "the public mode" isn't really descriptive.

src/ansi.rs
```diff
@@ -770,10 +778,104 @@ pub enum CursorShape {
...
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum PublicMode {
+    /// Known public mode.
+    Named(NamedPublicMode),
+    /// Unidentified publc mode.
+    Unidentified(u16),
```
chrisduerr, line 787: I think `Unknown` is the more common name for things like this and it's also shorter.

src/ansi.rs
```diff
@@ -770,10 +778,104 @@ pub enum CursorShape {
...
+
+impl PublicMode {
+    fn new(mode: u16) -> Self {
+        let named = match mode {
+            4 => NamedPublicMode::Insert,
+            20 => NamedPublicMode::LineFeedNewLine,
+            _ => return Self::Unidentified(mode),
+        };
+
+        Self::Named(named)
```
chrisduerr, line 798: Seems cleaner to just inline, that way control flow is simpler.

src/ansi.rs
```diff
@@ -770,10 +778,104 @@ pub enum CursorShape {
...
+            Self::Unidentified(mode) => mode,
+        }
+    }
+}
+
+impl From<NamedPrivateMode> for PrivateMode {
+    fn from(value: NamedPrivateMode) -> Self {
+        Self::Named(value)
+    }
+}
```
chrisduerr, line 874: Is this actually used?

chrisduerr (changes requested):
All comments for private also apply for public and vice-versa of course.

src/ansi.rs
kchibisov: Inside alacritty.

src/ansi.rs
```diff
@@ -119,7 +119,7 @@ impl Mul<f32> for Rgb {
             b: (f32::from(self.b) * rhs).clamp(0.0, 255.0) as u8,
         };
 
-        trace!("Scaling RGB by {} from {:?} to {:?}", rhs, self, result);
+        log::trace!("Scaling RGB by {} from {:?} to {:?}", rhs, self, result);
```
chrisduerr, line 122: This change confuses me. What's the motivation behind it?

src/ansi.rs
```diff
@@ -773,10 +781,100 @@ pub enum CursorShape {
...
+    }
+}
+
+/// ANSI modes.
+#[repr(u16)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum NamedMode {
+    /// IRM Insert Mode.
+    Insert = 4,
+    /// 20.
```
chrisduerr, line 823:
```suggestion
```

src/ansi.rs
```diff
@@ -1516,11 +1548,13 @@ where
             },
             ('L', []) => handler.insert_blank_lines(next_param_or(1) as usize),
             ('l', intermediates) => {
+                let is_private = intermediates.first() == Some(&b'?');
```
chrisduerr, line 1551: Wouldn't `('l', [b'?', ...])` work here? That way we could just use two separate match statements instead of having to create a separate conditional.

chrisduerr (changes requested):

src/ansi.rs
kchibisov: To remove `cfg` because `trace` is not used under some feature.

src/ansi.rs
kchibisov: it'll work just fine.

chrisduerr (approved):

PR #111, opened 2024-01-25
Add licenses to vte_generate_state_changes

sadmac7000:
Without these here, these don't end up in the crate tarball for this subcrate, making it hard for us to consume the crate from a legal standpoint.


kchibisov:
Just symlink the thing like it's done in other crates here.

kchibisov:
Ah, it's actually not symlinked, so better to make all of them symlink to root licenses...

sadmac7000:
> Ah, it's actually not symlinked in this repo, so better to make all of them symlink to root licenses...

Done.

kchibisov (approved):

PR #112, opened 2024-03-02
Support SCP control (SELECT CHARACTER PATH)

MoSal:
Modern usage of this control function comes from the BiDi draft proposal:

 https://terminal-wg.pages.freedesktop.org/bidi/recommendation/escape-sequences.html

 The draft slightly extends the definition in ECMA-48, as commented.

### DIFF
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -677,6 +677,9 @@ pub trait Handler {
     ///
     /// The output is of form `CSI > 4 ; mode m`.
     fn report_modify_other_keys(&mut self) {}
+
+    // Set SCP control.
+    fn set_scp(&mut self, _char_path: ScpCharPath, _update_mode: ScpUpdateMode) {}
 }
 
 bitflags! {
@@ -1195,6 +1198,32 @@ impl StandardCharset {
     }
 }
 
+/// SCP control's first parameter which determines character path.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ScpCharPath {
+    /// SCP's first parameter value of 0. Behavior is implementation defined.
+    Default,
+    /// SCP's first parameter value of 1 which sets character path to LEFT-TO-RIGHT.
+    LTR,
+    /// SCP's first parameter value of 2 which sets character path to RIGHT-TO-LEFT.
+    RTL,
+}
+
+/// SCP control's second parameter which determines update mode/direction between components.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ScpUpdateMode {
+    /// SCP's second parameter value of 0 (the default). Implementation dependant update.
+    ImplementationDependant,
+    /// SCP's second parameter value of 1.
+    ///
+    /// Reflect data component changes in the presentation component.
+    DataToPresentation,
+    /// SCP's second parameter value of 2.
+    ///
+    /// Reflect presentation component changes in the data component.
+    PresentationToData,
+}
+
 impl<'a, H, T> crate::Perform for Performer<'a, H, T>
 where
     H: Handler + 'a,
@@ -1550,6 +1579,30 @@ where
                 };
 
                 handler.clear_line(mode);
+            },
+            ('k', [b' ']) => {
+                // SCP control.
+                let char_path = match next_param_or(0) {
+                    0 => ScpCharPath::Default,
+                    1 => ScpCharPath::LTR,
+                    2 => ScpCharPath::RTL,
+                    _ => {
+                        unhandled!();
+                        return;
+                    },
+                };
+
+                let update_mode = match next_param_or(0) {
+                    0 => ScpUpdateMode::ImplementationDependant,
+                    1 => ScpUpdateMode::DataToPresentation,
+                    2 => ScpUpdateMode::PresentationToData,
+                    _ => {
+                        unhandled!();
+                        return;
+                    },
+                };
+
+                handler.set_scp(char_path, update_mode);
             },
             ('L', []) => handler.insert_blank_lines(next_param_or(1) as usize),
             ('l', []) => {


chrisduerr (changes requested):
Why are you interested in adding this protocol?

kchibisov:
Also, terminal-wg is long dead, and most proposals from it were not adopted or were altered later by downstream, so adding stuff from it is a bit questionable.

MoSal:
> Why are you interested in adding this protocol?

I plan to add support for the BiDi draft in `alacritty_terminal` (hidden behind a feature of course), then utilize that to add full support in `cosmic-term`.

Will use forks if necessary. No problem.

MoSal:
> Also, terminal-wg is long dead, and most proposals from it were not adopted or were altered later by downstream, so adding stuff from it is a bit questionable.

I'm not aware of any *modern* attempts at standardizing BiDi support in terminals.

The draft is still implemented by VTE-based (the Gnome one) terminals. I've been testing against MATE Terminal to confirm.

Not sure what downstream you're referring to.

src/ansi.rs
```diff
@@ -1551,6 +1554,28 @@ where
...
+                // SCP control
+                //
+                // Changes from ECMA-48 as proposed by the BiDi draft proposal:
+                //  * Recognize value 0 for Ps1
+                //  * Allow passing no params or Ps1 alone
+                //  * Assume default 0 value for both Ps1 and Ps2
+                //
+                //  https://terminal-wg.pages.freedesktop.org/bidi/recommendation/escape-sequences.html
+                let ps1 = params_iter.next().unwrap_or(&[0]);
+                let ps2 = params_iter.next().unwrap_or(&[0]);
```
chrisduerr, line 1567:
```suggestion
                let ps1 = next_param_or(0);
                let ps2 = next_param_or(0);
```

src/ansi.rs
```diff
@@ -677,6 +677,9 @@ pub trait Handler {
     ///
     /// The output is of form `CSI > 4 ; mode m`.
     fn report_modify_other_keys(&mut self) {}
+
+    // Set SCP control
+    fn set_scp(&mut self, _ps1: u16, _ps2: u16) {}
```
chrisduerr, line 682: `ps1` and `ps2` are not appropriate variable names. Instead the actual purpose of the variables should be used. This applies to the whole PR.

src/ansi.rs
```diff
@@ -677,6 +677,9 @@ pub trait Handler {
     ///
     /// The output is of form `CSI > 4 ; mode m`.
     fn report_modify_other_keys(&mut self) {}
+
+    // Set SCP control
```
chrisduerr, line 681:
```suggestion
    // Set SCP control.
```

src/ansi.rs
```diff
@@ -1551,6 +1554,28 @@ where
 
                 handler.clear_line(mode);
             },
+            ('k', [b' ']) => {
+                // SCP control
+                //
+                // Changes from ECMA-48 as proposed by the BiDi draft proposal:
+                //  * Recognize value 0 for Ps1
```
chrisduerr, line 1561: What purpose does this have for applications? Having a right-to-left "default" seems unfeasible to me. So effectively this would just be another way to use `1`.

src/ansi.rs
```diff
@@ -1551,6 +1554,28 @@ where
...
+                //  * Allow passing no params or Ps1 alone
```
chrisduerr, line 1562: Is this actually used by applications? I don't really mind it but I don't see an advantage in applications supporting a newer "spec" that is much more likely to cause only partial support.

src/ansi.rs
```diff
@@ -1551,6 +1554,28 @@ where
...
+                //  * Assume default 0 value for both Ps1 and Ps2
+                //
+                //  https://terminal-wg.pages.freedesktop.org/bidi/recommendation/escape-sequences.html
+                let ps1 = params_iter.next().unwrap_or(&[0]);
+                let ps2 = params_iter.next().unwrap_or(&[0]);
+                if params_iter.next().is_some() {
+                    unhandled!();
```
chrisduerr, line 1569: Not exactly wrong, but looking at other escapes we have been mostly ignoring additional parameters.

Might be nice in case someone comes along and adds a new parameter to do whatever they seem is necessary, so we still support it without having to explicitly handle that parameter.

src/ansi.rs
```diff
@@ -1551,6 +1554,28 @@ where
...
+                } else {
+                    match (ps1, ps2) {
+                        (&[ps1], &[ps2]) if ps1 <= 2 && ps2 <= 2 => {
```
chrisduerr, line 1572: Arguably this is a restriction that doesn't benefit the consumer of VTE, since they can always deny values out of range themselves but once we deny them, they can never handle them.

Quickly looking at other escapes, it seems we mostly only error on unknown values when we're mapping to enums, which kinda makes sense to me. So I'd do the same here: Either map in an enum, or just pass values. Enums could be nice for documentation purposes.

chrisduerr (changes requested):
Sorry for the delay, I've kinda been avoiding this PR because I'm not a huge fan of the escape.

Considering that this is part of ECMA-48 and rather simple, I'm willing to include it in vte for downstream despite not being interested in it myself. However this would assume it actually gets used, since I'd rather not add stuff just because it exists.

The "terminal-wg" stuff can be largely ignored I think. Some of its stuff falls under valid implementation-dependent behavior for ECMA-48 anyway.

MoSal:
> Sorry for the delay

Ditto.

> However this would assume it actually gets used, since I'd rather not add stuff just because it exists.

Agreed.

For me personally, if [support in `alacritty_terminal`](https://github.com/MoSal/alacritty/commit/b4f8e72e616fd751eb51fd440c8cbdbcd0401239) will not be accepted, then there is no point in including this in vte upstream.

chrisduerr:
> For me personally, if support in alacritty_terminal will not be accepted, then there is no point in including this in vte upstream.

Thanks for pointing this out. The patch doesn't look **too** bad, but it would have a massive performance impact so there's no way it could be accepted for upstream usage. Effectively just making it something we have to carry around for another terminal emulator. This just looks like a draft so I don't want to be too hasty in outright shooting it down, but I'm struggling to justify it from Alacritty's perspective.

@kchibisov Any thoughts?

MoSal:
> but it would have a massive performance impact so there's no way it could be accepted for upstream usage.

Note that there is zero code change if the feature is not enabled.

kchibisov:
>would have a massive performance impact so there's no way it could be accepted for upstream usage

I think it'll be fine as long as it's in the extra data? Though, we might slightly tune it to reduce the boilerplate to write cow stuff correctly.

The patch doesn't touch the rest of code path, so I don't see how it'll hurt. Also, I'd rather measure perf with our bot to talk about perf penalties.

chrisduerr:
> I think it'll be fine as long as it's in the extra data? Though, we might slightly tune it to reduce the boilerplate to write cow stuff correctly.

Yeah that's probably what I would recommend for making it more feasible.

> Note that there is zero code change if the feature is not enabled.

Yes, this is what I meant by "upstream usage", as in actually supporting this in Alacritty. However even for `alacritty_terminal` I'm not sure I'd want something that actually impacts performance significantly without good reason. While this crate's goals certainly are a little less defined than Alacritty itself, I would still like to retain the principle that using alacritty_terminal should always result in a fast emulator (so it would need to go into extra data anyway).

> The patch doesn't touch the rest of code path, so I don't see how it'll hurt. Also, I'd rather measure perf with our bot to talk about perf penalties.

Yeah that's true. It's just something I noticed in the patch and I don't need the bot to know that putting anything in the cell data will impact performance. Just so @MoSal is aware.

MoSal:
I'm not sure what kind of performance regression should I be looking out for.

I ran vtebench in alacritty, first time built with defaults, then with the `bidi_draft` feature enabled. And I'm not seeing a clear performance regression.

A test case that would show such a regression would be most helpful.

both

kchibisov:
Your system is rather noise, the thing that regressed, I guess is that the cell size got increased, which we should never do.

MoSal:
> the thing that regressed, I guess is that the cell size got increased, which we should never do.

That, I get.

Unless I misunderstood, what's being suggested is moving `bidi_flags` to `CellExtra` like [this](https://github.com/MoSal/alacritty/commit/9494e733b8585302e12bb2d769f501b2daf396c6), right?

kchibisov:
>Unless I misunderstood, what's being suggested is moving bidi_flags to CellExtra like [this](https://github.com/MoSal/alacritty/commit/9494e733b8585302e12bb2d769f501b2daf396c6), right?

right, because the escape is fairly rare.

src/ansi.rs
```diff
@@ -1195,6 +1198,41 @@ impl StandardCharset {
     }
 }
 
+/// SCP control's first parameter which determines character path.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ScpCharPath {
+    /// SCP's first parameter value of 0 (the default). Behavior is implementation defined.
+    ///
+    /// This value is defined by Terminals Working Group BiDi draft proposal.
```
chrisduerr, line 1206: I don't really like this, considering `Terminals Working Group` is a dead project. A dead project's draft proposal isn't really the authority for anything and probably shouldn't be recommended in docs for developers new to the topic.

Behavior is best based on existing implementations and no reference to documentation might be better than fragile ones?

src/ansi.rs
```diff
@@ -1195,6 +1198,41 @@ impl StandardCharset {
...
+    Default,
+    /// SCP's first parameter value of 1 which sets character path to LEFT-TO-RIGHT.
+    LTR,
+    /// SCP's first parameter value of 2 which sets character path to RIGHT-TO-LEFT.
+    RTL,
+    /// SCP's first parameter with an unknown value.
+    Unknown(u16),
```
chrisduerr, line 1213: I don't think this is actually valid, right? I don't think we should give terminal emulators the option to implement new proprietary extensions.

src/ansi.rs
```diff
@@ -1195,6 +1198,41 @@ impl StandardCharset {
...
+    /// SCP's second parameter value of 1.
+    ///
+    /// Reflect data component changes in the presentation component.
+    DataToPresentation,
+    /// SCP's second parameter value of 2.
+    ///
+    /// Reflect presentation component changes in the data component.
+    PresentationToData,
+    /// SCP's second parameter with an unknown value.
+    Unknown(u16),
```
chrisduerr, line 1233: See the other unknown comment.

chrisduerr (changes requested):
This new code is clean enough to where I am willing to merge it as-is regardless of alacritty_terminal support as long as the remaining comments are addressed.

MoSal:
I think all review concerns are addressed.

chrisduerr (approved):

PR #110, opened 2024-01-21
Added APC support

John-Toohey:
Fixes #109.

This adds support for APC (Application Program Command) in the same way OSC is implemented.  It does, however, have the drawback of (just over) doubling the in-memory footprint of the `table::STATE_CHANGES` static, for it became necessary to split `State` and `Action` as the amount of variants within `Action` exceeded 15, meaning it was no longer possible for it to be a pseudo `u4`.

More detail in the commit messages.

Thank you for your time.

PR #115, opened 2024-08-04
Add support for custom parsing of APC, SOS and PM sequences.

### LINKED ISSUE #109: Support for APC?
I am trying to use `vte` to parse APC commands, which it seems to ignore by setting the state to `SosPmApcString` and then `Anywhere` through the duration of the command.  Would it be possible to adapt the `Parser` interface to support parsing these commands?

boazy:
Fixes #109.

This attempts the same goal at [John-Toohey](https://github.com/John-Toohey)'s PR #110, but since this PR was was not accepted, I wanted to try another approach and see whether it is acceptable.

## Rationale for support

APC sequences are rarely supported by general-purpose terminals (if we put aside Kermit clients, tmux and screen), but there is one exception: The [Kitty Terminal Graphics Protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol/). While the Kitty Image Protocol is not as ubiquitous as Sixel, it is more robust, accurate and powerful and it's implemented by several prominent terminal emulators:

* Kitty (obviously)
* Konsole
* Wezterm
* Wayst
* Ghostty

It's also supported by a large number of programs and libraries, some of them are mentioned [here](https://sw.kovidgoyal.net/kitty/graphics-protocol/).

### Design Goals

- Support SOS, PM and APC sequences (just like the original PR)
- Support streaming parsing of SOS, PM and APC sequence payloads.
- Avoid adding new conditional branches in existing code paths. This should minimize performance impact on existing code.
- Avoid re-purposing OSC state variables (in order to reduce complexity)
- Avoid adding any new stored state variable
- No embedded parsers - use existing state change tables for parsing

#### Reorder actions into packable and non-packable actions

Currently, resulting state and actions are packed into an 8-bit value in the state change table (with a 4-bit nibble allocated for each). All values from 0 to 15 for both state and actions are used, so I could not add new states and action that would be _packable_. Unfortunately, I needed to support a state change with the `OpaquePut` action. The best way I've found is to separate the integer values of actions that do not need to be packed into the state change table (such as `Hook, `Unhook` and `Clear`) into a value higher than 15 and reserve lower values for actions that need to be packed

I'm not sure if my current solution is acceptable or not but I have no other idea on how to resolve this situation without repurposing the `OscString` state and actions and re-introducing an embedded parser.

#### Action matching order

This should have minimal impact, but in order to avoid any performance impact for programs which do not use APC sequences, I made sure that when matching APC/SOS/PM-specific actions, they are always matched last.

#### Streaming

Unlike #110, this PR does not collect the sequence payload inside an array. Instead, the sequence payload is streamed directly to the `Perform` trait, one byte at a time. I believe this approach is more flexible and would lead to better performance overall. The key gains here are:

- Avoiding extra vector allocation for the payload (I don't want to re-use `osc_raw` and in any case we would have to increase its default size to 4096 to be compatible with the max payload size).
- Avoiding unnecessary copying when. The payload often just needs to be parsed directly. This is especially useful when sending images directly, since the image data would often just be forwarded to an image format parser anyway.

#### Names

I've changed `SosPmApcString` to `OpaqueString`, since I wanted to have a general name for all these types of sequences where the VTE parser has no understanding of the internal structure of the sequence payload. This is difference from OSC and CSI sequences where the VTE parser is aware of the high-level structure of the sequence (if not the semantics).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -864,36 +864,35 @@ pub trait Perform {
     /// encountered.
     fn sos_start(&mut self) {}
 
-    /// Invoked when the beginning of a new APC (Application Program Command)
-    /// sequence is encountered.
-    fn apc_start(&mut self) {}
+    /// Invoked for every valid byte (0x20-0xFF) in a SOS (Start of String)
+    /// sequence.
+    fn sos_put(&mut self, _byte: u8) {}
+
+    /// Invoked when the end of an SOS (Start of String) sequence is
+    /// encountered.
+    fn sos_end(&mut self) {}
 
     /// Invoked when the beginning of a new PM (Privacy Message) sequence is
     /// encountered.
     fn pm_start(&mut self) {}
 
-    /// Invoked for every valid character (0x20-0xFF) in a SOS (Start of String)
+    /// Invoked for every valid byte (0x20-0xFF) in a PM (Privacy Message)
     /// sequence.
-    fn sos_dispatch(&mut self, _byte: u8) {}
-
-    /// Invoked for every valid character (0x20-0xFF) in an APC (Application
-    /// Program Command) sequence.
-    fn apc_dispatch(&mut self, _byte: u8) {}
-
-    /// Invoked for every valid character (0x20-0xFF) in a PM (Privacy Message)
-    /// sequence.
-    fn pm_dispatch(&mut self, _byte: u8) {}
-
-    /// Invoked when the end of an SOS (Start of String) sequence is
-    /// encountered.
-    fn sos_string_end(&mut self) {}
-
-    /// Invoked when the end of an APC (Application Program Command) sequence is
-    /// encountered.
-    fn apc_string_end(&mut self) {}
+    fn pm_put(&mut self, _byte: u8) {}
 
     /// Invoked when the end of a PM (Privacy Message) sequence is encountered.
-    fn pm_string_end(&mut self) {}
+    fn pm_end(&mut self) {}
+
+    /// Invoked when the beginning of a new APC (Application Program Command)
+    /// sequence is encountered.
+    fn apc_start(&mut self) {}
+
+    /// Invoked for every valid byte (0x20-0xFF) in an APC (Application Program
+    /// Command) sequence.
+    fn apc_put(&mut self, _byte: u8) {}
+    /// Invoked when the end of an APC (Application Program Command) sequence is
+    /// encountered.
+    fn apc_end(&mut self) {}
 
     /// Whether the parser should terminate prematurely.
     ///
@@ -925,12 +924,12 @@ impl<P: Perform> OpaqueDispatch for SosDispatch<'_, P> {
 
     #[inline(always)]
     fn opaque_put(&mut self, byte: u8) {
-        self.0.sos_dispatch(byte);
+        self.0.sos_put(byte);
     }
 
     #[inline(always)]
     fn opaque_end(&mut self) {
-        self.0.sos_string_end();
+        self.0.sos_end();
     }
 }
 
@@ -944,12 +943,12 @@ impl<P: Perform> OpaqueDispatch for ApcDispatch<'_, P> {
 
     #[inline(always)]
     fn opaque_put(&mut self, byte: u8) {
-        self.0.apc_dispatch(byte);
+        self.0.apc_put(byte);
     }
 
     #[inline(always)]
     fn opaque_end(&mut self) {
-        self.0.apc_string_end();
+        self.0.apc_end();
     }
 }
 
@@ -963,12 +962,12 @@ impl<P: Perform> OpaqueDispatch for PmDispatch<'_, P> {
 
     #[inline(always)]
     fn opaque_put(&mut self, byte: u8) {
-        self.0.pm_dispatch(byte);
+        self.0.pm_put(byte);
     }
 
     #[inline(always)]
     fn opaque_end(&mut self) {
-        self.0.pm_string_end();
+        self.0.pm_end();
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -179,7 +179,9 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
             State::Escape => self.advance_esc(performer, byte),
             State::EscapeIntermediate => self.advance_esc_intermediate(performer, byte),
             State::OscString => self.advance_osc_string(performer, byte),
-            State::SosPmApcString => self.anywhere(performer, byte),
+            State::SosString => self.advance_opaque_string(SosDispatch(performer), byte),
+            State::ApcString => self.advance_opaque_string(ApcDispatch(performer), byte),
+            State::PmString => self.advance_opaque_string(PmDispatch(performer), byte),
             State::Ground => unreachable!(),
         }
     }
@@ -356,7 +358,12 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
                 performer.esc_dispatch(self.intermediates(), self.ignoring, byte);
                 self.state = State::Ground
             },
-            0x58 => self.state = State::SosPmApcString,
+            0x58 => {
+                self.state = {
+                    performer.sos_start();
+                    State::SosString
+                }
+            },
             0x59..=0x5A => {
                 performer.esc_dispatch(self.intermediates(), self.ignoring, byte);
                 self.state = State::Ground
@@ -374,7 +381,14 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
                 self.osc_num_params = 0;
                 self.state = State::OscString
             },
-            0x5E..=0x5F => self.state = State::SosPmApcString,
+            0x5E => {
+                performer.pm_start();
+                self.state = State::PmString
+            },
+            0x5F => {
+                performer.apc_start();
+                self.state = State::ApcString
+            },
             0x60..=0x7E => {
                 performer.esc_dispatch(self.intermediates(), self.ignoring, byte);
                 self.state = State::Ground
@@ -431,6 +445,41 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
                 self.action_osc_put_param()
             },
             _ => self.action_osc_put(byte),
+        }
+    }
+
+    #[inline(always)]
+    fn advance_opaque_string<D: OpaqueDispatch>(&mut self, mut dispatcher: D, byte: u8) {
+        match byte {
+            0x07 => {
+                // The standard only supports ST-terminated SOS/APC/PM strings, using either
+                // ESC-ST (ESC-\) and C1-ST (0x9C), but kitty (and probably some other
+                // terminals) also support bell-terminated strings. Some
+                // terminals (including Kitty), do not support C1-ST (0x9C) as a
+                // terminator, which means every character from 0x20-0xFF can be
+                // used with this sequence in theory.
+                dispatcher.opaque_end();
+                self.state = State::Ground
+            },
+            0x18 | 0x1A => {
+                // XTerm terminates SOS/APC/PM strings on C1 CAN (^X) and SUB (^Z). This is also
+                // the same behavior we implement for OSC strings.
+                dispatcher.opaque_end();
+                dispatcher.execute(byte);
+                self.state = State::Ground
+            },
+            0x1B => {
+                // Any escape code ends the SOS/APC/PM string. This is not standard behavior,
+                // but avoids having to keep additional state.
+                dispatcher.opaque_end();
+                self.state = State::Escape
+            },
+            0x20..=0xFF => {
+                // Only dispatch valid characters.
+                dispatcher.opaque_put(byte)
+            },
+            // Ignore all other control codes
+            _ => (),
         }
     }
 
@@ -743,7 +792,9 @@ enum State {
     Escape,
     EscapeIntermediate,
     OscString,
-    SosPmApcString,
+    SosString,
+    ApcString,
+    PmString,
     #[default]
     Ground,
 }
@@ -811,6 +862,41 @@ pub trait Perform {
     /// subsequent characters were ignored.
     fn esc_dispatch(&mut self, _intermediates: &[u8], _ignore: bool, _byte: u8) {}
 
+    /// Invoked when the beginning of a new SOS (Start of String) sequence is
+    /// encountered.
+    fn sos_start(&mut self) {}
+
+    /// Invoked when the beginning of a new APC (Application Program Command)
+    /// sequence is encountered.
+    fn apc_start(&mut self) {}
+
+    /// Invoked when the beginning of a new PM (Privacy Message) sequence is
+    /// encountered.
+    fn pm_start(&mut self) {}
+
+    /// Invoked for every valid character (0x20-0xFF) in a SOS (Start of String)
+    /// sequence.
+    fn sos_dispatch(&mut self, _byte: u8) {}
+
+    /// Invoked for every valid character (0x20-0xFF) in an APC (Application
+    /// Program Command) sequence.
+    fn apc_dispatch(&mut self, _byte: u8) {}
+
+    /// Invoked for every valid character (0x20-0xFF) in a PM (Privacy Message)
+    /// sequence.
+    fn pm_dispatch(&mut self, _byte: u8) {}
+
+    /// Invoked when the end of an SOS (Start of String) sequence is
+    /// encountered.
+    fn sos_string_end(&mut self) {}
+
+    /// Invoked when the end of an APC (Application Program Command) sequence is
+    /// encountered.
+    fn apc_string_end(&mut self) {}
+
+    /// Invoked when the end of a PM (Privacy Message) sequence is encountered.
+    fn pm_string_end(&mut self) {}
+
     /// Whether the parser should terminate prematurely.
     ///
     /// This can be used in conjunction with
@@ -822,6 +908,69 @@ pub trait Perform {
     #[inline(always)]
     fn terminated(&self) -> bool {
         false
+    }
+}
+
+trait OpaqueDispatch {
+    fn execute(&mut self, byte: u8);
+    fn opaque_put(&mut self, byte: u8);
+    fn opaque_end(&mut self);
+}
+
+struct SosDispatch<'a, P: Perform>(&'a mut P);
+
+impl<P: Perform> OpaqueDispatch for SosDispatch<'_, P> {
+    #[inline(always)]
+    fn execute(&mut self, byte: u8) {
+        self.0.execute(byte);
+    }
+
+    #[inline(always)]
+    fn opaque_put(&mut self, byte: u8) {
+        self.0.sos_dispatch(byte);
+    }
+
+    #[inline(always)]
+    fn opaque_end(&mut self) {
+        self.0.sos_string_end();
+    }
+}
+
+struct ApcDispatch<'a, P: Perform>(&'a mut P);
+
+impl<P: Perform> OpaqueDispatch for ApcDispatch<'_, P> {
+    #[inline(always)]
+    fn execute(&mut self, byte: u8) {
+        self.0.execute(byte);
+    }
+
+    #[inline(always)]
+    fn opaque_put(&mut self, byte: u8) {
+        self.0.apc_dispatch(byte);
+    }
+
+    #[inline(always)]
+    fn opaque_end(&mut self) {
+        self.0.apc_string_end();
+    }
+}
+
+struct PmDispatch<'a, P: Perform>(&'a mut P);
+
+impl<P: Perform> OpaqueDispatch for PmDispatch<'_, P> {
+    #[inline(always)]
+    fn execute(&mut self, byte: u8) {
+        self.0.execute(byte);
+    }
+
+    #[inline(always)]
+    fn opaque_put(&mut self, byte: u8) {
+        self.0.pm_dispatch(byte);
+    }
+
+    #[inline(always)]
+    fn opaque_end(&mut self) {
+        self.0.pm_string_end();
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -450,33 +450,20 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
     fn advance_opaque_string<D: OpaqueDispatch>(&mut self, mut dispatcher: D, byte: u8) {
         match byte {
             0x07 => {
-                // The standard only supports ST-terminated SOS/APC/PM strings, using either
-                // ESC-ST (ESC-\) and C1-ST (0x9C), but kitty (and probably some other
-                // terminals) also support bell-terminated strings. Some
-                // terminals (including Kitty), do not support C1-ST (0x9C) as a
-                // terminator, which means every character from 0x20-0xFF can be
-                // used with this sequence in theory.
                 dispatcher.opaque_end();
                 self.state = State::Ground
             },
             0x18 | 0x1A => {
-                // XTerm terminates SOS/APC/PM strings on C1 CAN (^X) and SUB (^Z). This is also
-                // the same behavior we implement for OSC strings.
                 dispatcher.opaque_end();
                 dispatcher.execute(byte);
                 self.state = State::Ground
             },
             0x1B => {
-                // Any escape code ends the SOS/APC/PM string. This is not standard behavior,
-                // but avoids having to keep additional state.
                 dispatcher.opaque_end();
                 self.state = State::Escape
             },
-            0x20..=0xFF => {
-                // Only dispatch valid characters.
-                dispatcher.opaque_put(byte)
-            },
-            // Ignore all other control codes
+            0x20..=0xFF => dispatcher.opaque_put(byte),
+            // Ignore all other control bytes.
             _ => (),
         }
     }
@@ -908,6 +895,10 @@ pub trait Perform {
     }
 }
 
+/// This trait is used internally to provide a common implementation for Opaque
+/// Sequences (SOS, APC, PM). Implementations of this trait will just forward
+/// calls to the equivalent method on [Perform]. Implementations of this trait
+/// are always inlined to avoid overhead.
 trait OpaqueDispatch {
     fn execute(&mut self, byte: u8);
     fn opaque_put(&mut self, byte: u8);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -359,10 +359,8 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
                 self.state = State::Ground
             },
             0x58 => {
-                self.state = {
-                    performer.sos_start();
-                    State::SosString
-                }
+                performer.sos_start();
+                self.state = State::SosString
             },
             0x59..=0x5A => {
                 performer.esc_dispatch(self.intermediates(), self.ignoring, byte);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -979,12 +979,21 @@ mod tests {
         b'c', b'r', b'i', b't', b't', b'y', 0x07, // End OSC
     ];
 
+    const ST_ESC_SEQUENCE: &[Sequence] = &[Sequence::Esc(vec![], false, 0x5C)];
+
     #[derive(Default)]
     struct Dispatcher {
         dispatched: Vec<Sequence>,
     }
 
-    #[derive(Debug, PartialEq, Eq)]
+    #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+    enum OpaqueSequenceKind {
+        Sos,
+        Pm,
+        Apc,
+    }
+
+    #[derive(Clone, Debug, PartialEq, Eq)]
     enum Sequence {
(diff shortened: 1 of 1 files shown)


chrisduerr (commented):
I'll give a detailed review later today/tomorrow, but assuming that this was benchmarked it seems like a reasonable approach.

Generally I think the VTE crate needs a little rewrite, but I don't want to hold up other changes for that since it will likely take a long time before I get around to that.

boazy:
> I'll give a detailed review later today/tomorrow, but assuming that this was benchmarked it seems like a reasonable approach.

I ran `cargo bench` but I'm a bit unsure on how to run the full vtebench suite. Is there a guide for that?

I don't expect any performance impact since I've added no new state and I've only added tailed branches to an already-long match expression, but you never know until you test...

kchibisov:
You can open a PR against alacritty with your changes, so we can use our bot for it. But generally you just build alacritty with your changes and run the `vtebench`.

boazy:
> You can open a PR against alacritty with your changes, so we can use our bot for it. But generally you just build alacritty with your changes and run the `vtebench`.

Thanks! Here are the results:

Master Branch:
```
  dense_cells (625 samples @ 1 MiB):
    15.47ms avg (90% < 16ms) +-1.93ms

  scrolling (199 samples @ 1 MiB):
    37.19ms avg (90% < 42ms) +-4.62ms

  scrolling_bottom_region (361 samples @ 1 MiB):
    27.25ms avg (90% < 28ms) +-2.28ms

  scrolling_bottom_small_region (369 samples @ 1 MiB):
    26.69ms avg (90% < 27ms) +-2.3ms

  scrolling_fullscreen (185 samples @ 1 MiB):
    41.15ms avg (90% < 45ms) +-2.28ms

  scrolling_top_region (247 samples @ 1 MiB):
    40.09ms avg (90% < 42ms) +-2.92ms

  scrolling_top_small_region (364 samples @ 1 MiB):
    27.06ms avg (90% < 28ms) +-2.27ms

  unicode (611 samples @ 1.06 MiB):
    15.86ms avg (90% < 17ms) +-1.55ms
```

This PR:
```
  dense_cells (606 samples @ 1 MiB):
    16ms avg (90% < 17ms) +-1.26ms

  scrolling (198 samples @ 1 MiB):
    37.39ms avg (90% < 42ms) +-5.06ms

  scrolling_bottom_region (361 samples @ 1 MiB):
    27.25ms avg (90% < 28ms) +-2.12ms

  scrolling_bottom_small_region (362 samples @ 1 MiB):
    27.19ms avg (90% < 28ms) +-2.25ms

  scrolling_fullscreen (182 samples @ 1 MiB):
    41.63ms avg (90% < 45ms) +-3.56ms

  scrolling_top_region (250 samples @ 1 MiB):
    39.6ms avg (90% < 41ms) +-2.85ms

  scrolling_top_small_region (362 samples @ 1 MiB):
    27.16ms avg (90% < 28ms) +-1.28ms

  unicode (654 samples @ 1.06 MiB):
    14.79ms avg (90% < 16ms) +-2.14ms
```

The `scrolling_*` benchmarks are unstable and give me different lead every time. Unicode seems to be consistently slightly faster on this PR while `dense_cells` is slightly faster on `master` branch, but the difference is quite small and the variance is higher.

kchibisov:
@boazy just open a PR against alacritty's repo, it should _better_ for us to review that way, since we can run the bot.

boazy:
@kchibisov I've added a pull request. I'm not sure how to trigger the benchmarks bot.

chrisduerr:
> I'm not sure how to trigger the benchmarks bot.

You can't. I did.

src/definitions.rs
```diff
@@ -28,21 +28,34 @@ pub enum State {
...
+    Ignore = 5,
+    OscPut = 6,
+    Param = 7,
+    Print = 8,
+    Put = 9,
+    BeginUtf8 = 10,
+    OpaquePut = 11,
+
+    // Actions that do not need to be packed as 4 bits in the state table
+    // Can have values higher than 16
```
chrisduerr, line 44:
```suggestion
    // Actions that do not need to be packed as 4 bits in the state table
    // can have values higher than 16.
```

src/lib.rs
```diff
@@ -364,6 +371,24 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
             Action::BeginUtf8 => self.process_utf8(performer, byte),
             Action::Ignore => (),
             Action::None => (),
+
+            // APC Actions are checked last, since they are relatively rare
```
chrisduerr, line 375:
```suggestion
            // APC Actions are checked last, since they are relatively rare.
```

src/lib.rs
```diff
@@ -428,6 +453,20 @@ pub trait Perform {
...
+    ///
+    /// Until the opaque sequence ends (at which point `opaque_end` will be called), invalid
+    /// characters will be ignored while valid characters will be passed on to `opaque_put`.
+    fn opaque_start(&mut self, _kind: OpaqueSequenceKind) {}
+
+    /// We've reached the end of the ongoing opaque sequence (SOS, PM or APC).
+    fn opaque_end(&mut self) {}
+
+    /// A byte has been received as part of an ongoing opaque sequence.
+    fn opaque_put(&mut self, _byte: u8) {}
```
chrisduerr, line 469: Is there a good reason why you decided to make this one set of functions with a `_kind` parameter, instead of one for each kind?

chrisduerr (changes requested):

src/lib.rs
boazy: There are several reasons for this:

1. I wanted to minimize the amount of packed actions that I would need to add, since the current parser only supports 15 packed actions. If I separate these functions, we will need to track states for APC, SOS and PM sequences separately and then `Action::OpaquePut` would need to be packed, so this would consume two extra packed action slots. I wanted to minimize future performance impact for parsers that don't need to deal with these sequences, so I opted to unify all opaque sequence handling.
(`opaque_start` and `opaque_end` won't consume a slot, but I thought it's better to keep the names consistent)

2. SOS and PM sequences are even more rare. I've only found mention of SOS used as a C1 control code for MARC-21 Unicode variant. It is not even used in the context of a terminal or a terminal multiplexer.
PM is used by [screen](https://www.gnu.org/software/screen/manual/html_node/Privacy-Message.html) for displaying a notification and by [jexer](https://gitlab.com/AutumnMeowMeow/jexer/-/wikis/jexer-sequences#nested-text-mouse-pointer) for handling nested mouse pointers. I wasn't sure if it's a good idea to dedicate separate functions for these rare use cases.

3. APC, SOS and PM all work the same way[^1]. They all take opaque strings that are terminated with ST and they all don't define any semantics for how these strings should work. Since there is no well-defined structure, the implementation needs to create its own parser and track its own state for these strings anyway. In that case, separating APC, SOS and PM to different sets of functions doesn't help the application very much.

---

That's more or less my rationale, but I'm open to hear what you think about it.

[^1]: I'm stretching the truth a bit: According to the ECMA-48 spec, SOS allows any character to be embedded (except for escape codes or C1 sequences representing SOS or ST), while PM and APC only allow the ranges 0x08-0x0d, 0x20-0x7e. In practice, since nobody (as far as I know) is using SOS sequences as far as we know, it's quite hard for me to know how to implement it properly.

chrisduerr: If you pass them as an enum argument to this function, then this means you need to identify them anyway. It just seems like a way to avoid doing it properly and if we're not doing it properly what's the point? A type of escape sequence should either be **properly** supported, or not at all.

boazy: >  If you pass them as an enum argument to this function, then this means you need to identify them anyway.

This was my third point. You need to identify the actual (private) sequence type that APC/PM/SOS is used for anyway, even if I separate the function.

But I understand that you see separating them out as more proper.

What would you expect then? Three sets of functions and adding the necessary extra packed states?

chrisduerr: I would expect three sets of functions. How that is done is an implementation detail irrelevant to the consumer (as long as it's fast).

chrisduerr:
@boazy This has stalled out for a while now, are you still interested in working on this? I'm slightly concerned because the longer we wait, the higher the chance that a vte rewrite will make all your work worthless.

boazy:
> I would expect three sets of functions. How that is done is an implementation detail irrelevant to the consumer (as long as it's fast).

> @boazy This has stalled out for a while now, are you still interested in working on this? I'm slightly concerned because the longer we wait, the higher the chance that a vte rewrite will make all your work worthless.

I didn't have time for a while, but I've had some time again today and I've checked your suggestion. Unfortunately, I don't think it would be possible without completely re-architecting the way state works. As I've mentioned before, we only have 4 bits for actions that change the state, but we also only have 4 bits for the state itself, and here all possible values (0-15) are used.

If we want to support three sets of functions, we would need to track three separate states for each type of string (SOS, PM, APC). This is not possible without either:
1. Adding a new state variable for tracking the type of opaque string. (with possible performance impact)
2. Modifying the state variable to a u16 or better yet something like `struct ResultingState(State, Action)`.
Would this impact performance? I don't know, but it's also a big change I'm a little bit uneasy with...

The only thing I can do without any major change is to have 3 different functions instead of `opaque_start`, but `opaue_end` and `opaque_put` would have to remain a single function in this case.

boazy:
> But couldn't you just store that on `Parser` (the size of which is irrelevant), then branch inside `Action::OpaquePut =>` and `Action::OpaqueEnd =>` since doing so should leave performance of the other branches unaffected?
> 
> I believe that's kinda your first suggestion? Why do you think this would impact performance?

I honestly don't know. I assumed the size is important for performance since we want to be able to have all state transitions in one table (which only has 256 entries?). If we go back to the start, I created this PR trying to answer the comments to the original PR that tried to add APC support (#110), particularly this comment:

>  This basically feels like an embedded parser. Most of our parser is structured in a way to stream through bytes and process them as they're coming in, however rather than creating state transitions you're storing a state transition to later then match on it instead. VTE is already too slow and I'm afraid this will just amplify that problem.

https://github.com/alacritty/vte/pull/110#discussion_r1545286222

I realize I can add additional state that doesn't affect OSC sequences, but seeing this comment I was a bit worried. Do you believe adding an additional field to `Parser` (let's say `opaque_sequence_kind`) should be ok?

chrisduerr:
> I honestly don't know. I assumed the size is important for performance since we want to be able to have all state transitions in one table (which only has 256 entries?). If we go back to the start, I created this PR trying to answer the comments to the original PR that tried to add APC support (#110), particularly this comment:

Yeah that's fair. This would make this a slow sloppy implementation and it might not be worth adding it for that reason. But that applies to your current solution too and the only alternative is a bigger rework that integrates this as a core part of our parser.

> I realize I can add additional state that doesn't affect OSC sequences, but seeing this comment I was a bit worried. Do you believe adding an additional field to Parser (let's say opaque_sequence_kind) should be ok?

I think we should either do this, or go for something that properly integrates everything into our standard processing pipeline, which of course is difficult without affecting performance. I'm very hesitant to accept this sort of "second rate escape sequence", but I would prefer that over the current implementation.

boazy:
> I think we should either do this, or go for something that properly integrates everything into our standard processing pipeline, which of course is difficult without affecting performance. I'm very hesitant to accept this sort of "second rate escape sequence", but I would prefer that over the current implementation.

Ok, I think I've understood you now. I'll modify the pull request along these lines.

src/lib.rs
```diff
@@ -385,10 +386,34 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
-                performer.opaque_start(kind)
-            },
-            Action::OpaquePut => performer.opaque_put(byte),
-            Action::OpaqueEnd => performer.opaque_end(),
+                self.opaque_sequence_kind = Some(kind);
+                match kind {
+                    OpaqueSequenceKind::Sos => performer.sos_start(),
+                    OpaqueSequenceKind::Pm => performer.pm_start(),
+                    OpaqueSequenceKind::Apc => performer.apc_start(),
+                }
```
chrisduerr, line 394: The kind is assigned earlier in a match statement on the byte. It seems pointless to immediately match here again.
Should just call the functions in the first match.

chrisduerr (changes requested):

chrisduerr:
Going it tenatively approve the current revision, I don't think there's any other changes necessary from your end.

While I'm not entirely convinced I'll give this a closer look and if there are no performance regressions it should be fine.

chrisduerr:
@boazy Just a heads-up: Once https://github.com/alacritty/vte/pull/118 is merged, it might make some sense to reinvestigate this. There's certainly been some changes and we do have a lot of extra free states/actions that could be made use of without affecting performance.

heysweet:
Hello! Just wanted to re-ping this thread to see what progress was being here. We'd really like to support Kitty Image protocol in Warp, and this PR would help us get there

boazy:
This PR was languishing for a while now and now, and now that vte is ready, I probably need to rewrite it I guess, but I didn't have time to revisit this recently. I'll try to find time to do it.

The good news is that it should impact performance. I hope.

heysweet:
Hello again, sorry to be a bother, just wanted to check in on the status of this. We’re kicking off the full-time effort to support kitty image protocol starting this week, and just want to get an understanding how much of a risk this PR is to our project. Thank you!

boazy:
@chrisduerr , @kchibisov I've finally got time to rebase the PR to work with the new version. I haven't written any unit tests yet, since I want to see if you are ok with the direction.

Some points to take note of:
1. I'm currently not keeping any extra state for opaque sequence parsing, but that is introducing some limitations, which I will describe below.
2. There is no clear specification on how to handle any character besides 0x20-0x7F, 0x9C (C1 ST) and the ESC ST sequence (0x1B 0x5C) in APC/SOS/PM sequences. The original DEC VT* terminals seem to have ignored these sequences completely, so essentially everything until the ST sequence (which put the terminal back in ground state) was ignored.
3. What characters are valid inside an opaque sequence are also not clear and different terminals have different implementations. I went with supporting any non-control character, but that means that C1 ST (0x9C) currently does not terminate an opaque sequence. I didn't check all terminals who implement APC for Kitty protocol features, but Kitty and Ghostty do not seem to support C1 ST as well. In any case, I believe we should only implement C1 ST support if we support the C1 APC/SOS/PM codes for *starting* the sequence (which vte does not support), so this behavior is probably correct.
5. On the other hand, the opaque sequence can be terminated with non-standard bell (0x07), since kitty supports bell-terminated sequences, and the main usage for these sequences seem to be implementing Kitty protocol features anyway).
6. It is unclear how to handle escape sequences other than ST (`0x1b 0x5c`) inside an SOS/APC/PM string. My current approach is to terminate the sequence on any escape code, just as we vte does with OSC sequences, but this is not the most correct approach. A better approach would be to ignore these sequences, but the requires keeping more state.
7. I've tried to inline everything, and avoid both code duplication and unnecessary branching (by using the OpaqueDispatch trait) wherever possible.

There are two approaches I can see for keeping track of additional state (particularly important for item #6):
1. Add 3 additional states: `SOSEscape`, `APCEscape` and `PMEscape`. I will add special `advance_esc` implementation for each of these states, which ignore any other ESC sequence apart from `ESC-\` (ST).
2. Add an additional field to the parser that specifies the current opaque sequence mode (or none if no such mode exists). This field will be checked every time we handle an escape sequence, so it may impact the performance of non-SOS/APC/PM usage.

I believe the first approach should be better.

@heysweet Sorry for the time it took me to rebase this PR, I basically had to rewrite it from scratch.

j4james:
> On the other hand, the opaque sequence can be terminated with non-standard bell (0x07)

Note that most terminals only accept `BEL` as a terminator for `OSC` sequences, strictly for backwards compatibility. In other string sequences it's just another data character. Kitty only started accepting it as a terminator for all string sequences a couple of years ago - I don't know why. Personally I don't think it's a good idea to encourage non-standard usage like that.

boazy:
One more point of concern:
`[sos/apc/pm]_dispatch()` is called individually for each byte. For any use case which is not a byte-by-byte parser, it would probably be more efficient to call it on byte slices instead (I believe this was the entire point of #118).
The main issue with calling this on byte slices, is that the SOS/APC/PM sequences can be arbitrarily long; indeed, with the the most common use case (Kitty Image Protocol) these sequences could easily be several megabytes long. I am a little bit unsure how accumulating these sequence in a `Vec<u8>` (like what the parser is doing for OSC) would impact performance.

j4james:
> with the the most common use case (Kitty Image Protocol) these sequences could easily be several megabytes long.

Assuming apps actually follow the [spec](https://sw.kovidgoyal.net/kitty/graphics-protocol/#remote-client), then kitty image sequences should never be more than 4K ("the pixel data must first be base64 encoded then chunked up into chunks no larger than 4096 bytes"). But otherwise you're right - in theory these string sequences can be infinitely large.

boazy:
> > with the the most common use case (Kitty Image Protocol) these sequences could easily be several megabytes long.
> 
> Assuming apps actually follow the [spec](https://sw.kovidgoyal.net/kitty/graphics-protocol/#remote-client), then kitty image sequences should never be more than 4K ("the pixel data must first be base64 encoded then chunked up into chunks no larger than 4096 bytes"). But otherwise you're right - in theory these string sequences can be infinitely large.

Makes sense. I wasn't aware of the actual Kitty Image Protocol spec so I've missed that.
VTE could also implement its own chunking and make that configurable in the parser.

kchibisov:
The perf looks pretty much the same https://github.com/alacritty/alacritty/pull/8507

Can not really answer wrt design now questions you've raised.

src/lib.rs
```diff
@@ -811,6 +862,41 @@ pub trait Perform {
...
+    /// Invoked when the beginning of a new APC (Application Program Command)
+    /// sequence is encountered.
+    fn apc_start(&mut self) {}
+
+    /// Invoked when the beginning of a new PM (Privacy Message) sequence is
+    /// encountered.
+    fn pm_start(&mut self) {}
+
+    /// Invoked for every valid character (0x20-0xFF) in a SOS (Start of String)
+    /// sequence.
```
chrisduerr, line 878:
```suggestion
    /// Invoked for every byte (0x20-0xFF) in a SOS (Start of String) sequence.
```

Same comment applies for the other functions. Calling these "valid characters" could be misleading to consumers.

src/lib.rs
```diff
@@ -811,6 +862,41 @@ pub trait Perform {
...
+    /// encountered.
+    fn sos_start(&mut self) {}
+
+    /// Invoked when the beginning of a new APC (Application Program Command)
+    /// sequence is encountered.
+    fn apc_start(&mut self) {}
+
+    /// Invoked when the beginning of a new PM (Privacy Message) sequence is
+    /// encountered.
+    fn pm_start(&mut self) {}
```
chrisduerr, line 875: Rather than sorting these by start/put/end, I'd prefer grouping them by sos/apc/pm. I think it makes more sense since the grouping into start/put/end is somewhat arbitrary considering they aren't really interconnected.

src/lib.rs
```diff
@@ -811,6 +862,41 @@ pub trait Perform {
...
+    /// sequence.
+    fn sos_dispatch(&mut self, _byte: u8) {}
+
+    /// Invoked for every valid character (0x20-0xFF) in an APC (Application
+    /// Program Command) sequence.
+    fn apc_dispatch(&mut self, _byte: u8) {}
+
+    /// Invoked for every valid character (0x20-0xFF) in a PM (Privacy Message)
+    /// sequence.
+    fn pm_dispatch(&mut self, _byte: u8) {}
```
chrisduerr, line 887: I'm not sure the naming `_dispatch` is consistent with the rest of our methods. The existing `_dispatch` functions generally represent the dispatch of an entire escape sequence in full, while this function just represents the dispatch of a single byte in the sequence.

I think this is much closer to the way esc works in VTE, where it is split in `hook`, `put`, and `unhook`.

Other parts of this patch already make use of the `_put` nomenclature, so I think it's better to be consistent and rename these functions to `sos/pm/apc_put`.

src/lib.rs
```diff
@@ -825,6 +911,69 @@ pub trait Perform {
     }
 }
 
+trait OpaqueDispatch {
+    fn execute(&mut self, byte: u8);
+    fn opaque_put(&mut self, byte: u8);
+    fn opaque_end(&mut self);
+}
```
chrisduerr, line 918: Honestly this trait and its implementations is the main issue I have with this patch. It should work fine since it all just gets inlined and optimized out essentially, but it's still a whole lot of boilerplate.

I'm not sure I have any better ideas for now, but at the very least this trait needs a comment explaining that it's just a helper for dispatching over the opaque string escapes and in practice is inlined everywhere to function as static dispatch without conditional indirection.

chrisduerr (changes requested):
Honestly seems fine to me. Just some minor nits.

chrisduerr:
Just to be clear on this:

> [sos/apc/pm]_dispatch() is called individually for each byte. For any use case which is not a byte-by-byte parser, it would probably be more efficient to call it on byte slices instead (I believe this was the entire point of #118).
The main issue with calling this on byte slices, is that the SOS/APC/PM sequences can be arbitrarily long; indeed, with the the most common use case (Kitty Image Protocol) these sequences could easily be several megabytes long. I am a little bit unsure how accumulating these sequence in a Vec<u8> (like what the parser is doing for OSC) would impact performance.

I personally don't mind this "simplistic" approach because I have very little interest in sos/pm/apc escape sequences, so this way there is only a minimal amount of impact on Alacritty while VTE is kept simple. That's also why I don't care if these escapes are dispatched particularly efficiently or not, since ideally nobody every uses them.

That said, the collection into an (Array)Vec in VTE is essentially a solved issue. We have this for OSCs already where things are somewhat dynamically limited and having a matching/similar approach probably wouldn't be a huge deal for these escapes either. Ideally to ensure parser state doesn't get too big one would likely reuse the same byte buffers for both, since it's not possible to have two different escape sequences at the same time anyway.

But going for something like that is likely to slow down the implementation process, since it doesn't sound like that's something you're particularly familiar with. So I don't mind just going with the status quo in this patch.

src/lib.rs
```diff
@@ -434,6 +448,41 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+                // terminals) also support bell-terminated strings. Some
+                // terminals (including Kitty), do not support C1-ST (0x9C) as a
+                // terminator, which means every character from 0x20-0xFF can be
+                // used with this sequence in theory.
+                dispatcher.opaque_end();
+                self.state = State::Ground
+            },
+            0x18 | 0x1A => {
+                // XTerm terminates SOS/APC/PM strings on C1 CAN (^X) and SUB (^Z). This is also
+                // the same behavior we implement for OSC strings.
```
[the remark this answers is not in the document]
boazy: I was a bit confused, since not all terminals seem to do that (at least when handling SOS/PM/APC), but I'll remove the comment.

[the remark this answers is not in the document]
chrisduerr: Yeah I'll refer to https://vt100.net/emu/dec_ansi_parser for "general" guidance on Alacritty's parser here. It's not really a SOS/PM/APC, but more of an "anywhere" thing. As such a SOS/PM/APC-specific comment is unnecessary.

src/lib.rs
```diff
@@ -434,6 +448,41 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+                dispatcher.opaque_end();
+                dispatcher.execute(byte);
+                self.state = State::Ground
+            },
+            0x1B => {
+                // Any escape code ends the SOS/APC/PM string. This is not standard behavior,
+                // but avoids having to keep additional state.
```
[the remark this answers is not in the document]
boazy: I'm sorry if I got my intentions wrong. The comments are not meant to be here to stay. They are mostly documenting my thoughts while looking for feedback to this pull request. I have little experience with de-facto terminal implementations, so I can only refer to the specifications (which are a bit vague).
If escape sequences resetting the state is common de-facto behavior, then I would be happy to keep it this way, since it keeps the implementation clean and simple.

src/lib.rs
```diff
@@ -434,6 +448,41 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
         }
     }
 
+    #[inline(always)]
+    fn advance_opaque_string<D: OpaqueDispatch>(&mut self, mut dispatcher: D, byte: u8) {
+        match byte {
+            0x07 => {
+                // The standard only supports ST-terminated SOS/APC/PM strings, using either
+                // ESC-ST (ESC-\) and C1-ST (0x9C), but kitty (and probably some other
+                // terminals) also support bell-terminated strings. Some
```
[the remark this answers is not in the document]
boazy: I was referring to SOS/PM/APC sequences, which are not yet supported by alacritty. The current behavior (treating them as the _anywhere_ state) doesn't seem to support termination with BEL, so this would be a behavior change for alacrity, wouldn't it?

But since as you've said, this is how OSC sequences already behave, so I think we should change this behavior to be more consistent.

As stated above, I added these comments only to explain my reasoning, I didn't intend them to be included in the code in the final version of the PR.

src/lib.rs
```diff
@@ -434,6 +448,41 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+            },
+            0x1B => {
+                // Any escape code ends the SOS/APC/PM string. This is not standard behavior,
+                // but avoids having to keep additional state.
+                dispatcher.opaque_end();
+                self.state = State::Escape
+            },
+            0x20..=0xFF => {
+                // Only dispatch valid characters.
+                dispatcher.opaque_put(byte)
```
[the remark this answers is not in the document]
boazy: Sorry for that. This is a leftover comment from a previous implementation attempt that I gave up on and forgot to remove.

src/lib.rs
boazy: Makes sense. I'll do it.

src/lib.rs
boazy: You're absolutely right, considering the method on the internal `OpaqueDispatch` trait is called `opaque_put` and then it ends up calling `sos/apc/pm_dispatch()`...

src/lib.rs
boazy: This trait was the best solution I could think of until now, which means it is the least terrible one. I'll add a comment.

boazy:
> That said, the collection into an (Array)Vec in VTE is essentially a solved issue. We have this for OSCs already where things are somewhat dynamically limited and having a matching/similar approach probably wouldn't be a huge deal for these escapes either. Ideally to ensure parser state doesn't get too big one would likely reuse the same byte buffers for both, since it's not possible to have two different escape sequences at the same time anyway.

Familiarity is not my issue here. I do understand how to collect bytes into a Vec (I hope). I thought of reusing `osc_raw`, but there may be several (minor) performance implications for Alacritty (and any other user of Parser which ignores SOS/PM/APC sequences):

* If SOS/PM/APC sequences do get sent by a program, they'll be written to a buffer and then just get ignored later. In theory, this would impact performance.
* Kitty Image Protocol sequences are still bigger than typical APM sequences, even with 4096 byte chunking (together with all the parameters that come before the payload, the actual chunk size would be a bit larger than 4096 bytes). If any program sends these sequences by mistake, the `Vec` would grow to that size and stay there (the non-std implementation will not be impacted, if we keep `MAX_OSC_RAW` the same).

I don't think these two issues are big in practice. Programs that send Kitty Image Protocol sequences to a terminal that doesn't support that are already misbehaving, and 4096 or 8192 bytes of extra memory per terminal are not a big thing. I also don't think cache locality would be impacted, since the shorter OSC sequences would keep using just the beginning of the buffer.

I will create another branch and a draft PR with a buffered version for comparison.

boazy:
On a second thought, I'm not sure if buffering opaque sequences will be such a great boon for clients of VTE anyway. Just like VTE itself is inlining methods, clients can just choose to inline their `apc_put()` implementation and put everything inside a Vec. So I think I would be okay with this implementation.

chrisduerr:
> On a second thought, I'm not sure if buffering opaque sequences will be such a great boon for clients of VTE anyway. Just like VTE itself is inlining methods, clients can just choose to inline their apc_put() implementation and put everything inside a Vec. So I think I would be okay with this implementation.

If the goal was to write the *ideal* implementation as far as performance goes, we'd probably special-case the OSC state to look for its termination sequence using `memchr`. However while that would speed up opaque strings, it would probably reduce the non-opaque string escape parsing slightly, so I don't think it's worth it.

src/lib.rs
```diff
@@ -1386,6 +1571,103 @@ mod tests {
         }
     }
 
+    fn expect_opaque_sequence(
+        input: &[u8],
+        kind: OpaqueSequenceKind,
+        expected_payload: &[u8],
+        expected_trailer: &[Sequence],
```
chrisduerr, line 1578: This can just be a boolean to simplify this function and remove the necessity for a constant. `st_terminated: bool` or something like that.

src/lib.rs
```diff
@@ -1386,6 +1571,103 @@ mod tests {
...
+    ) {
+        let mut expected_dispatched: Vec<Sequence> = vec![Sequence::OpaqueStart(kind)];
```
chrisduerr, line 1580:
```suggestion
        let mut expected_dispatched = vec![Sequence::OpaqueStart(kind)];
```

Is this really necessary? I'd be extremely surprised if Rust wasn't able to infer it here.

chrisduerr:
Added some small comments, but didn't want to actually do a review yet. Please just re-request one once you think you're done with your changes.

chrisduerr:
Closing due to inactivity.

PR #116, opened 2024-10-21
OSS-Fuzz: Add fuzzers for OSS-Fuzz integration

arthurscchan:
This PR creates a cargo fuzz directory along with a fuzzing harness. The aim is to include the utf8parse module for fuzzing under OSS-Fuzz.

Fuzzing is essentially a stress testing technique used to uncover bugs in software, and OSS-Fuzz is a free service run by Google that provides continuous fuzzing for important open-source projects. Integrating utf8parse would be beneficial in identifying potential memory corruption issues.

The only requirement for integration at this stage is an email linked to a Google account, which will be used to receive notifications when bugs are detected.

An initial integration has been submitted via https://github.com/google/oss-fuzz/pull/12627, which relies on the fuzzer being hosted upstream. This is the primary objective of this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,4 +30,4 @@ no_std = ["arrayvec"]
 serde = ["dep:serde"]
 
 [workspace]
-members = ["utf8parse", "vte_generate_state_changes"]
+members = ["utf8parse", "vte_generate_state_changes", "utf8parse/fuzz"]
--- a/utf8parse/fuzz/.gitignore
+++ b/utf8parse/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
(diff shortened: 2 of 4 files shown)


chrisduerr (changes requested):
VTE is already being fuzz tested and this PR just tests the least interesting bit of the crate.

Honestly this just looks like a low effort advertising attempt, which I can not support.

DavidKorczynski:
Hey @chrisduerr -- apologies for the impression, that was certainly not the intention.

We'd be happy to extend on the fuzzing to get a complete set up. Would you be interested in this?

With respect to the existing fuzzing being done, then is this publicly available? I saw a few PRs/Issues referencing AFL, but nothing in the code?

chrisduerr:
> With respect to the existing fuzzing being done, then is this publicly available?

No. Fuzzing VTE is so trivial, it's not necessary to have a specific setup clogging up the repository. That way people can use whatever tool they're most familiar with. Considering that the parser just takes a single byte as input, it could not be any simpler really.

PR #117, opened 2024-11-01
Fix clippy warnings

gwenn:


chrisduerr (changes requested):
No point in making separate commits for clippy changes. That just generates unnecessary noise and maintenance burden.

gwenn:
Would you mind explaining how these clippy warnings should be fixed ?
Or how do you work locally without this noise ?

PR #119, opened 2024-12-25
Add support for DECST8C to reset tab stops to default

aymanbagabas:
This adds support for the DECST8C sequence to reset tab stops to default 8 column stops.

See https://vt100.net/docs/vt510-rm/DECST8C.html


src/ansi.rs
```diff
@@ -1520,6 +1523,11 @@ where
             ('E', []) => handler.move_down_and_cr(next_param_or(1) as usize),
             ('F', []) => handler.move_up_and_cr(next_param_or(1) as usize),
             ('G', []) | ('`', []) => handler.goto_col(next_param_or(1) as usize - 1),
+            ('W', [b'?']) => {
+                if next_param_or(0) == 5 {
+                    handler.reset_tabs();
+                }
```
chrisduerr, line 1529: Needs unhandled in the `else` branch.

chrisduerr (changes requested):
Missing a changelog entry.

src/ansi.rs
aymanbagabas: Done

aymanbagabas:
> Missing a changelog entry.

Do you need me to add an entry to CHANGELOG.md? If so, should that be under version 0.14.0?

aymanbagabas:
> Missing a changelog entry.

Updated the changelogs

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+## 0.14.0
+
+- Support `CSI ? 5 W` to reset tabs stops to 8 columns
```
chrisduerr, line 6: I'll be able to give a proper review later today, but this changelog message will certainly have to change. It's not guaranteed that tab stops are going to be 8 columns wide. Should say something like "... to the default" or similar.

chrisduerr (changes requested):

CHANGELOG.md
j4james: Note that this escape sequences is [`DECST8C`](https://vt100.net/docs/vt510-rm/DECST8C.html), which is supposed to "set a tab stop at every eight columns starting with column 9" (quoting the VT510 and VT520/VT525 programmer manuals).  If that's not what it's doing, I'd think that's a problem with the implementation rather than the change log.

src/ansi.rs
```diff
@@ -566,6 +566,9 @@ pub trait Handler {
     /// Clear tab stops.
     fn clear_tabs(&mut self, _mode: TabulationClearMode) {}
 
+    /// Reset tab stops to default.
+    fn reset_tabs(&mut self) {}
```
chrisduerr, line 570: At the very least this needs to be changed to be correct. IIRC the only way to modify tab stops is based on current cursor position, so I think a fixed 'reset to 8' should be fine. But if there's an escape to set the tabstops at arbitrary position, it might be nice to have a generic function that just uses an interval to set the tabstops.

chrisduerr (changes requested):

src/ansi.rs
j4james: There is an escape sequence that allows you to set a list of individual tab stops at arbitrary positions (see [`DECRSPS`](https://vt100.net/docs/vt510-rm/DECRSPS.html)), but I'm not aware of anything that'll automatically set all tab stops at a given interval (other than `DECST8C`).

aymanbagabas: I don't see the need to make it generic. By default, terminals start with tab stops at every 8th column. This escape code simply resets the tab stops to the default.

If the need arises in the future, something like `reset_tabs_to(interval)` can be added.

kchibisov: wouldn't terminfo value for tabsize mess with the default, but not this escape? Though no one really uses other tabsizes these days.

chrisduerr: > By default, terminals start with tab stops at every 8th column.

This is not necessarily true.

aymanbagabas: > > By default, terminals start with tab stops at every 8th column.
> 
> This is not necessarily true.

Fair enough. I've updated the patch to accept an interval to set the tabstops

src/ansi.rs
```diff
@@ -612,6 +612,9 @@ pub trait Handler {
     /// Clear tab stops.
     fn clear_tabs(&mut self, _mode: TabulationClearMode) {}
 
+    /// Reset tab stops to every `interval` column.
+    fn reset_tabs(&mut self, _interval: u16) {}
```
chrisduerr, line 616: I'm not sure `reset` is helpful here, there's no guarantee it ever was 8 to begin with.

I think this function should either be `set_tabs/tabstops`, or something even more specific like setting it to every 8 columns.

src/ansi.rs
```diff
@@ -1574,6 +1577,14 @@ where
             ('E', []) => handler.move_down_and_cr(next_param_or(1) as usize),
             ('F', []) => handler.move_up_and_cr(next_param_or(1) as usize),
             ('G', []) | ('`', []) => handler.goto_col(next_param_or(1) as usize - 1),
+            ('W', [b'?']) => {
+                if next_param_or(0) == 5 {
+                    const DEFAULT_INTERVAL: u16 = 8;
+                    handler.reset_tabs(DEFAULT_INTERVAL);
```
chrisduerr, line 1583: The constant here doesn't really help and would never be changed. The escape itself specifies it as 8, so a terminal emulator with tabstops at every 16 columns should still set it to 8.

chrisduerr (changes requested):

src/ansi.rs
aymanbagabas: I've changed the function name to `set_tabs`

src/ansi.rs
aymanbagabas: Removed the constant and replaced it with a number

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+## 0.15.0
```
chrisduerr, line 4:
```suggestion
## Unreleased
```

Shouldn't make a changelog entry for a version that doesn't exist yet. And this isn't breaking anyway.

src/ansi.rs
```diff
@@ -1574,6 +1577,13 @@ where
             ('E', []) => handler.move_down_and_cr(next_param_or(1) as usize),
             ('F', []) => handler.move_up_and_cr(next_param_or(1) as usize),
             ('G', []) | ('`', []) => handler.goto_col(next_param_or(1) as usize - 1),
+            ('W', [b'?']) => {
+                if next_param_or(0) == 5 {
```
chrisduerr, line 1581:
```suggestion
            ('W', [b'?']) if next_params_or(0) == 5 => {
```

I think this should work and allow you to avoid explicitly writing the unhandled branch?

chrisduerr (changes requested):

src/ansi.rs
aymanbagabas: Done

chrisduerr (approved):

Issue #82, opened 2023-03-01
Misc clap needs


epage:
For clap and related applications, I've been looking into ANSI parsers.  None quite meet my needs so I was looking at writing my own (or forking vte and turning it into what I want) but I figured I'd reach out first in case there is a way to make this work within vte and its of interest.

vte seems like it is optimized for developers for use in alacritty.  Performance would most be noticed in highly interactive applications which would have a high escape code to text ratio.  Since every control code needs to be processed to render correctly for alacritty, vte cares more about the control code side of printable control codes like `\n`.  As the build times and binary size are likely a drop in bucket for alacritty, I'm assuming they haven't been optimized.

My care abouts are the opposite of the above.  My applications are static and likely to have a low escape code to text ratio and I would want to treat all printable control codes as text.  Technically, this can all be handled with vte's design but the char-by-char processing won't be the most optimal.  Instead I'd want to deal with slices of text.  `clap` is also a heavily used crate and there is a lot of interest in managing the build times and compile sizes.  I could likely get away without the proc macro generated state tables and replace them with a function with matches and have little performance hit but massive compile time and binary size improvements.

What I'm trying to decide is how much to contribute `vte` to handle both cases or if its better to go my own route.  Thoughts?

(sorry, wasn't there if there was a better medium to reach out)

chrisduerr:
I'm not sure how much details you actually need to figure out about the escape sequences, or if you're just looking to discard them, but to me it sounds like there's more things that your usecase doesn't share with `vte` than things that are shared.

Ultimately if you take out the char-by-char parsing, remove the performance, then throw out some of the escape sequences you don't care about, there's not much of an escape sequence parser left.

If you think you can make things work *without* affecting VTE's performance, I'd be happy to help adapting it to your usecase. But it sounds like that might not be the best approach depending on what your actual application is. That said, it might not make much sense to rewrite VTE entirely unless the char-by-char parsing performance and binary size actually make a difference, so I'd probably try and measure that first.

epage:
For build times, my initial changes to `vte` dropped debug build times for `parselog` from 1.5s to 0.5s which is significant for clap.   `proc-macro2` would be a reused dependency for `clap_derive` users which helps somewhat.

I'm actually surprised that switching away from the table didn't change binary size, so that could possibly be reintroduced with [self modifying code](https://matklad.github.io/2022/03/26/self-modifying-code.html).

[...]

PR #122, opened 2025-01-09
Bump version to 0.14.0

chrisduerr:


kchibisov (approved):

kchibisov:
Also, maybe state that utf8parse is dead?

chrisduerr:
> Also, maybe state that utf8parse is dead?

Hm that's a good question. Technically utf8parse is fine the way it is, it just won't get changed in the future easily.

I'm willing to wait until someone complains and if they do we can just create a separate repository.

jayvdb:
@kchibisov @chrisduerr
I see it was removed here https://github.com/alacritty/vte/commit/7321a442a6fc0fc5b6d6ed7af364477d25e706fd

https://github.com/trailofbits/cargo-unmaintained now reports

```
utf8parse (not in https://github.com/alacritty/vte)
```

I was in the process of raising an issue at https://github.com/rust-cli/anstyle and https://github.com/rustsec/advisory-db about utf8parse now being unmaintained.
If there is an option of this org creating a new repo for utf8parse and new release to update the crates.io metadata, that would be great.  Otherwise other projects need to find an alternative or create & maintain a fork.

chrisduerr:
@jayvdb I think I'd like to hold off on creating a repository until a maintainer actually using `utf8parse` comes forward and expresses an interest in the crate.

I'd argue that generally speaking the crate on its own shouldn't really provide much of value, but I don't know what `anstyle` is using it for. I'm happy to host it under the alacritty org, but it's never really been a crate I'd consider particularly 'maintained'.

jayvdb:
You can see how it is being used fairly quickly by looking at this recent rejected PR https://github.com/rust-cli/anstyle/pull/219

utf8parse has a *lot* of users, esp by via two anstyle crates; see https://crates.io/crates/utf8parse/reverse_dependencies

If it has never really been considered 'maintained' by the alacritty org, then perhaps it should be recreated in a different org, probably https://github.com/rust-cli where anstyle resides.

chrisduerr:
Anstyle's usage of this crate is exactly *why* I question whether this crate is necessary or not. Because I would argue it's not useful to anstyle.

kchibisov:
>If it has never really been considered 'maintained' by the alacritty org, then perhaps it should be recreated in a different org, probably https://github.com/rust-cli where anstyle resides.

It was a support crate for vte crate, once we were able to replace it with something else `std::str` in this case, we have removed it. From what I can see, utf8parse is only used by crates related to vte in the past. In general, std lib could be used instead of it, and I think even in old vte.

>I'd argue that generally speaking the crate on its own shouldn't really provide much of value, but I don't know what anstyle is using it for. I'm happy to host it under the alacritty org, but it's never really been a crate I'd consider particularly 'maintained'.

They forked from vte at some point, so I guess that's why it's like that.

chrisduerr:
> They forked from vte at some point, so I guess that's why it's like that.

Yeah exactly. They use it because VTE did, so VTE changing away from it because it's less efficient is likely a good time to reconsider the usage of utf8parse. Adopting the current parser would make a big difference especially for text-heavy data.

epage:
For context that led to the fork, see #82.

A big part of the fork was to drop the proc-macro which is now gone.

I also do operations on slices, rather than per-byte, as the encoding and decoding of a `char` can be expensive I have to have a buffer to hold all of this.  It seems vte now accepts slices but still only outputs a `char` at a time.

A nice to have would be for `Parser` to implement the various standard traits, like `Debug`.

I could dig into `vte`s new code to see what I can carry over to my fork and/or I can dig into alternative ways of dealing with `utf8parse`.  In both cases, its going to take investigation time when I am in the middle of quite a lot of other stuff.

chrisduerr:
> It seems vte now accepts slices but still only outputs a char at a time.

Just for context: This was done to avoid breaking VTE's API too much, since any terminal emulator will have to process char-by-char anyway. In theory printing strings wouldn't be a big change, it would just be a breaking one.

> A nice to have would be for Parser to implement the various standard traits, like Debug.

Standard traits should be implemented, though introspection into what the parser is doing internally is not something that really should matter to consumers of the library.

> I could dig into vtes new code to see what I can carry over to my fork and/or I can dig into alternative ways of dealing with utf8parse. In both cases, its going to take investigation time when I am in the middle of quite a lot of other stuff.

I don't think there's any rush here. The last version of utf8parse hasn't changed in a long time and will keep working just fine in the future. It's also been fuzzed multiple times so I'd be surprised about major issues with it.

That said, it doesn't really sound like keeping utf8parse around would be in your best interest, so I don't see a reason to keep it maintained personally.

epage:
Agreed that i don't see a reason to rush something.

To track slice support, I created #127.

PR #124, opened 2025-01-11
Rewrite table based state change to `match` based

kchibisov:
The table based state change was too complex to make guesses why it's getting slow and too fragile, as in modifying the amount of states/actions were slowing down, even though, they were not used.

Rewrite the state + action change exactly how it's in [1] with respect to our modifications/C1, etc. The new implementation is generally faster than the previous one and is easier for compiler to reason about and generate more efficient structures.

Also, the structure got way simpler to follow, since it matches the spec pretty much exactly.

[1] - https://vt100.net/emu/dec_ansi_parser


kchibisov:
Benchmark result https://github.com/alacritty/alacritty/pull/8426#issuecomment-2585373509

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
-                },
-                _ => (),
-            }
-        } else {
-            self.perform_action(performer, action, byte);
+    #[inline(always)]
+    fn advance_dcs_ignore<P: Perform>(&mut self, performer: &mut P, byte: u8) {
+        match byte {
+            0x00..=0x17 | 0x19 | 0x1C..=0x7F => (),
+            _ => self.anywhere(performer, byte),
```
chrisduerr, line 300: Technically we could just replace this entire method body with a call to `self.anywhere`.

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+            0x20..=0x2F => {
+                self.action_collect(byte);
+                self.state = State::DcsIntermediate
+            },
+            0x30..=0x39 => self.action_paramnext(byte),
+            0x3A => self.action_subparam(),
+            0x3B => self.action_param(),
+            0x3C..=0x3F => self.state = State::DcsIgnore,
+            0x40..=0x7E => self.action_hook(performer, byte),
+            0x7F => self.state = State::DcsIgnore,
```
chrisduerr: I think this should be ignore, rather than a DcsIgnore state transfer.

```suggestion
            0x7F => (),
```

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+                self.state = State::OscString
+            },
+            0x5E..=0x5F => self.state = State::SosPmApcString,
+            0x60..=0x7E => {
+                performer.esc_dispatch(self.intermediates(), self.ignoring, byte);
+                self.state = State::Ground
+            },
+            _ => self.anywhere(performer, byte),
+        }
+    }
```
chrisduerr, line 402: Generally I appreciate having everything laid out in byte order. But honestly this function would be half as long if we didn't do it.

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
-                );
-            },
-            Action::EscDispatch => {
-                performer.esc_dispatch(self.intermediates(), self.ignoring, byte);
+    #[inline(always)]
+    fn advance_sos_pm_apc_string<P: Perform>(&mut self, performer: &mut P, byte: u8) {
+        match byte {
+            0x00..=0x17 | 0x19 | 0x1C..=0x7F => (),
+            _ => self.anywhere(performer, byte),
+        }
```
chrisduerr, line 454: Could also just call self.anywhere here.

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+    fn action_param(&mut self) {
+        if self.params.is_full() {
+            self.ignoring = true;
+        } else {
+            self.params.push(self.param);
+            self.param = 0;
+        }
+    }
+
+    // Neither 0x3A nor 0x3B.
```
chrisduerr, line 527: I think it's better to document what the function does, rather than how it is entered.

In this case `paramnext` advances while inside a parameter, without terminating that parameter.

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+    fn action_subparam(&mut self) {
+        if self.params.is_full() {
+            self.ignoring = true;
+        } else {
+            self.params.extend(self.param);
+            self.param = 0;
+        }
+    }
+
+    // When we transition from the 0x3B.
```
chrisduerr, line 516: I think it's better to document what the function does, rather than how it is entered.

In this case `action_params` advances to the next parameter.

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+    fn action_collect(&mut self, byte: u8) {
+        if self.intermediate_idx == MAX_INTERMEDIATES {
+            self.ignoring = true;
+        } else {
+            self.intermediates[self.intermediate_idx] = byte;
+            self.intermediate_idx += 1;
+        }
+    }
+
+    // When we transition from the 0x3A.
```
chrisduerr, line 505: I think it's better to document what the function does, rather than how it is entered.
Can still have the characters in the comment, but honestly for this kind of stuff I'd use the plain character rather than the hex. `:` and `;` are much more clear.

In this case `action_params` advances to the next subparameter.

This can also just be doc comments, no reason not to.

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+        if self.params.is_full() {
+            self.ignoring = true;
+        } else {
+            // Continue collecting bytes into param
+            self.param = self.param.saturating_mul(10);
+            self.param = self.param.saturating_add((byte - b'0') as u16);
+        }
+    }
+
+    // Add OSC param separator
```
chrisduerr, line 539:
```suggestion
    /// Add OSC param separator.
```

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+    #[inline]
+    fn action_osc_put_param(&mut self) {
+        let idx = self.osc_raw.len();
+
+        let param_idx = self.osc_num_params;
+        match param_idx {
+            // First param is special - 0 to current byte index
```
chrisduerr, line 546:
```suggestion
            // First param is special - 0 to current byte index.
```

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+            0 => self.osc_params[param_idx] = (0, idx),
+
+            // Only process up to MAX_OSC_PARAMS
```
chrisduerr, line 549:
```suggestion
            // Only process up to MAX_OSC_PARAMS.
```

src/lib.rs
```diff
@@ -173,164 +194,388 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+            MAX_OSC_PARAMS => return,
+
+            // All other params depend on previous indexing
```
chrisduerr, line 552:
```suggestion
            // All other params depend on previous indexing.
```

chrisduerr (changes requested):

src/lib.rs
kchibisov: they are in byte order though here.

src/lib.rs
```diff
@@ -113,20 +110,17 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
-        if self.partial_utf8_len > 0 {
+        if self.partial_utf8_len != 0 {
             i += self.advance_partial_utf8(performer, bytes);
         }
 
         while i != bytes.len() {
             match self.state {
                 State::Ground => i += self.advance_ground(performer, &bytes[i..]),
                 _ => {
+                    // Do not inline.
```
chrisduerr, line 121: We should mention *why* here.

src/lib.rs
```diff
@@ -159,12 +153,9 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
```
chrisduerr, line 156: Same missing reason.

src/lib.rs
```diff
@@ -173,164 +164,398 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
...
+        }
+    }
+
+    /// Advance inside the parameter without terminating it.
+    #[inline]
+    fn action_paramnext(&mut self, byte: u8) {
+        if self.params.is_full() {
+            self.ignoring = true;
+        } else {
+            // Continue collecting bytes into param
```
chrisduerr, line 513:
```suggestion
            // Continue collecting bytes into param.
```

chrisduerr (changes requested):

chrisduerr (approved):

jwilm:
This is awesome! Nice work @kchibisov and @chrisduerr.

PR #125, opened 2025-01-12
Fix crash when valid char was split

kchibisov:
If the valid character was split across reads of partial utf8 and got terminated by invalid byte, we should print it and advance, instead of trying to discard it entirely.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -460,6 +460,22 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
                 c.len_utf8() - old_bytes
             },
             Err(err) => {
+                let valid_bytes = err.valid_up_to();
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -460,20 +460,20 @@ impl<const OSC_RAW_BUF_SIZE: usize> Parser<OSC_RAW_BUF_SIZE> {
                 c.len_utf8() - old_bytes
             },
             Err(err) => {
+                let valid_bytes = err.valid_up_to();
```
chrisduerr, line 463: I'd move the `if valid_bytes > 0` stuff here, should probably be the most hit branch anyway. Then either early return or else with a match that only has `to_copy` in none, doesn't really matter either way but  early return might look nicer wrt indentation.

chrisduerr (changes requested):

chrisduerr (approved):

PR #126, opened 2025-01-12
Bump version to 0.14.1

chrisduerr:


kchibisov (approved):
could also mention perf change.

chrisduerr:
> could also mention perf change.

One could, but I think it shouldn't really affect people much.

PR #129, opened 2025-01-27
Replace `no_std` with `std` feature

### LINKED ISSUE #128: `memchr` should be special handled for `no_std`
`memchr` has the `std` feature enabled by default, while `vte` does not disable this feature when the `no_std` feature is enabled, which result in a compiling failure for `vte` in a `no_std` environment.

chrisduerr:
The `no_std` feature has been broken since version `0.14`, since it did not disable the `std` feature of `memchr`.

To fall more in line with the general ecosystem standards, this crate changes the `no_std` feature into an `std` feature. Since `vte` should be significantly faster with this feature, it has also been enabled by default (previously `no_std` was default).

This is a breaking change, even if we ignore `no_std` consumers which aren't able to compile 0.14, since intermediate dependencies relying on the `no_std` feature will have to remove it from the manifest.

Closes #128.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ CHANGELOG
 ## Unreleased
 
 - Support `CSI ? 5 W` to reset tabs stops to every 8th column
+- Replaced `no_std` with a new `std` feature
+- Changed default features to include `std`
 
 ## 0.14.1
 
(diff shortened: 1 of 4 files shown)


chrisduerr:
I'm not sure if a breaking change for this is 'worth it', but I feel like it's beneficial long-term.

kchibisov (approved):
Given that features are generally additive, that's how it should have been.

PR #137, opened 2025-03-30
Activate ansi feature for docs.rs

gwenn:
Apply:
https://users.rust-lang.org/t/how-to-document-optional-features-in-api-docs/64577/3
And update build script accordingly
Currently, there is a warning:
```
warning: unresolved link to `Perform::advance`
   --> src/lib.rs:139:15
    |
139 |     /// See [`Perform::advance`] for more details.
    |               ^^^^^^^^^^^^^^^^ the trait `Perform` has no associated item named `advance`
    |
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: `vte` (lib doc) generated 1 warning
```
But I am not sure if the link should be on `Parser::advance` or `Perfom::terminated`.

### DIFF
--- a/.builds/linux.yml
+++ b/.builds/linux.yml
@@ -26,4 +26,4 @@ tasks:
   - rustdoc: |
       $HOME/.cargo/bin/rustup toolchain install nightly -c rust-docs
       cd vte
-      RUSTDOCFLAGS="--cfg docsrs -Dwarnings" $HOME/.cargo/bin/cargo +nightly doc --features=ansi --no-deps
+      RUSTDOCFLAGS="--cfg docsrs -Dwarnings" $HOME/.cargo/bin/cargo +nightly doc --all-features --no-deps
--- a/.builds/linux.yml
+++ b/.builds/linux.yml
@@ -24,6 +24,6 @@ tasks:
       rm Cargo.lock
       $HOME/.cargo/bin/cargo +$msrv test
   - rustdoc: |
-      $HOME/.cargo/bin/rustup toolchain install nightly -c rustdoc
+      $HOME/.cargo/bin/rustup toolchain install nightly -c rust-docs
       cd vte
       RUSTDOCFLAGS="--cfg docsrs -Dwarnings" $HOME/.cargo/bin/cargo +nightly doc --features=ansi --no-deps
--- a/.builds/linux.yml
+++ b/.builds/linux.yml
@@ -23,3 +23,7 @@ tasks:
       $HOME/.cargo/bin/rustup toolchain install --profile minimal $msrv
       rm Cargo.lock
       $HOME/.cargo/bin/cargo +$msrv test
+  - rustdoc: |
+      $HOME/.cargo/bin/rustup toolchain install nightly -c rustdoc
+      cd vte
+      RUSTDOCFLAGS="--cfg docsrs -Dwarnings" $HOME/.cargo/bin/cargo +nightly doc --features=ansi --no-deps
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,7 @@ cursor-icon = { version = "1.0.0", default-features = false, optional = true }
 log = { version = "0.4.17", optional = true }
 memchr = { version = "2.7.4", default-features = false }
 serde = { version = "1.0.160", features = ["derive"], optional = true }
+
+[package.metadata.docs.rs]
+features = ["ansi"]
+rustdoc-args = ["--cfg", "docsrs"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,5 +28,5 @@ memchr = { version = "2.7.4", default-features = false }
 serde = { version = "1.0.160", features = ["derive"], optional = true }
 
 [package.metadata.docs.rs]
-features = ["ansi"]
+all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,6 +28,7 @@
 //! [Paul Williams' ANSI parser state machine]: https://vt100.net/emu/dec_ansi_parser
 #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use)]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(docsrs, feature(doc_cfg))]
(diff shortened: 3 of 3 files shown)


chrisduerr:
> But I am not sure if the link should be on Parser::advance or Perfom::terminated.

Should be `Self::advance`.

chrisduerr (changes requested):
Seems reasonable, I'll take a proper look once CI passes.

Generally if there's other features that are doc relevant, they should also be considered for this patch though. No reason to limit it to ansi.

gwenn:
> Seems reasonable, I'll take a proper look once CI passes.

Done.
> Generally if there's other features that are doc relevant, they should also be considered for this patch though. No reason to limit it to ansi.

I see only `serde` but it doesn't seem relevant to me.

.builds/linux.yml
```diff
@@ -23,3 +23,7 @@ tasks:
       $HOME/.cargo/bin/rustup toolchain install --profile minimal $msrv
       rm Cargo.lock
       $HOME/.cargo/bin/cargo +$msrv test
+  - rustdoc: |
+      $HOME/.cargo/bin/rustup toolchain install nightly -c rust-docs
```
chrisduerr, line 27: Is there a reason why this needs to use nightly? Otherwise this should just be added to the clippy component install.

Cargo.toml
```diff
@@ -26,3 +26,7 @@ cursor-icon = { version = "1.0.0", default-features = false, optional = true }
 log = { version = "0.4.17", optional = true }
 memchr = { version = "2.7.4", default-features = false }
 serde = { version = "1.0.160", features = ["derive"], optional = true }
+
+[package.metadata.docs.rs]
+features = ["ansi"]
```
chrisduerr, line 31: Ansi isn't the only feature, so I'm not sure it's correct to provide only ansi here?

src/lib.rs
```diff
@@ -28,6 +28,7 @@
 //! [Paul Williams' ANSI parser state machine]: https://vt100.net/emu/dec_ansi_parser
 #![deny(clippy::all, clippy::if_not_else, clippy::enum_glob_use)]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
chrisduerr, line 31: Any reason not to use `doc_auto_cfg` instead? (https://github.com/rust-random/rand/blob/master/src/lib.rs#L56)

chrisduerr (changes requested):

.builds/linux.yml
gwenn:
```sh
$ RUSTDOCFLAGS="--cfg docsrs -Dwarnings" $HOME/.cargo/bin/cargo doc --features=ansi --no-deps
    Checking log v0.4.26
    Checking bitflags v2.9.0
    Checking cursor-icon v1.1.0
 Documenting vte v0.15.0 (C:\Users\treguiergw\perso\vte)
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> src\lib.rs:31:21
   |
31 | #![cfg_attr(docsrs, feature(doc_cfg))]
   |                     ^^^^^^^^^^^^^^^^

```

Cargo.toml
gwenn: Do you mean that we should add `serde` ?
https://github.com/alacritty/vte/pull/137#issuecomment-2768132829
> I see only serde but it doesn't seem relevant to me.

src/lib.rs
gwenn: Sorry, I didn't know!
https://doc.rust-lang.org/rustdoc/unstable-features.html?highlight=doc_a#doc_auto_cfg-automatically-generate-doccfg

I need to check.
Thanks.

nixpulvis: `#![doc(cfg_hide(doc))]` makes my head spin a bit. What a messy feature. Glad it works though.

Cargo.toml
chrisduerr: I'm not sure how these features are used, but if they're just listed on docs.rs, then all features should be included.

src/lib.rs
gwenn: > Any reason not to use `doc_auto_cfg` instead? (https://github.com/rust-random/rand/blob/master/src/lib.rs#L56)

https://github.com/alacritty/vte/pull/137/commits/027150de3314b330160fa2a69f8da36855c5379c

Cargo.toml
gwenn: > I'm not sure how these features are used, but if they're just listed on docs.rs, then all features should be included.

https://github.com/alacritty/vte/pull/137/commits/cba10b84882939079ebde2f18509afeeecabb4ba

.builds/linux.yml
```diff
@@ -23,3 +23,7 @@ tasks:
...
+      cd vte
+      RUSTDOCFLAGS="--cfg docsrs -Dwarnings" $HOME/.cargo/bin/cargo +nightly doc --features=ansi --no-deps
```
chrisduerr, line 29: Is there any reason to have the `--cfg docsrs` here? What exactly does it achieve? I'd assume rustdoc verification should be largely independent of target documentation platform.

chrisduerr (changes requested):

.builds/linux.yml
gwenn: I am just following:
https://users.rust-lang.org/t/how-to-document-optional-features-in-api-docs/64577/3

 You gave me a link to rand crate:
 >  Any reason not to use doc_auto_cfg instead? (https://github.com/rust-random/rand/blob/master/src/lib.rs#L56)
 
 And it seems that they do the same:
 https://github.com/rust-random/rand/blob/7808f4e1f7d9578e3c08405b1f4e575720d58a51/.github/workflows/test.yml#L37
!

kchibisov: >I'd assume rustdoc verification should be largely independent of target documentation platform.

unless you have doc stuff behind `docsrs` feature, it shouldn't really matter.

gwenn: I am almost sure that we can use any name instead of `docsrs`.
Because, it is used only to activate the nightly `doc_auto_cfg` feature.

chrisduerr (approved):

PR #139, opened 2025-04-02
add test for params mod

ericwq:
add unit test for Debug trait


chrisduerr (changes requested):
Completely unnecessary.

ericwq:
at least it improve the coverage.

PR #143, opened 2025-08-02
Add Clone+Debug derives to vte::Parser

zhiburt:
I wonder whether it's reasonable?

Take care.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,7 +52,7 @@ const MAX_OSC_RAW: usize = 1024;
 ///
 /// Generic over the value for the size of the raw Operating System Command
 /// buffer. Only used when the `std` feature is not enabled.
-#[derive(Default)]
+#[derive(Debug, Default, Clone)]
 pub struct Parser<const OSC_RAW_BUF_SIZE: usize = MAX_OSC_RAW> {
     state: State,
     intermediates: [u8; MAX_INTERMEDIATES],
--- a/src/params.rs
+++ b/src/params.rs
@@ -4,7 +4,7 @@ use core::fmt::{self, Debug, Formatter};
 
 pub(crate) const MAX_PARAMS: usize = 32;
 
-#[derive(Default)]
+#[derive(Default, Clone)]
 pub struct Params {
     /// Number of subparameters for each parameter.
     ///


chrisduerr:
Is this something you actually need? While adding these isn't necessarily bad, I don't see why someone would need to use them.

nixpulvis:
I mean, if you don't need to clone the parser, you can always implement `Clone` yourself and just skip that part (initialize a new parser). The real question is if you need to preserve the state of the parser, and what value cloning the parser actually has.

As for `Debug`, I generally think everything should implement this in one form or another unless there's a good reason not to.

zhiburt:
> The real question is if you need to preserve the state of the parser, and what value cloning the parser actually has.

Exactly;
That's why it's impossible to implement `Clone` in client code.

> As for Debug, I generally think everything should implement this in one form or another unless there's a good reason not to.

Same

chrisduerr:
> As far as I understand it won't hurt as it won't be compiled if not used.

That's not entirely accurate because you're still dealing with the compilation overhead. It's effectively compiled and then removed again. It also makes it much easier to *accidentally* use it.

> As for Debug, I generally think everything should implement this in one form or another unless there's a good reason not to.

Well the automatically derived impl certainly isn't the way to go, that'll just spill out a bunch of internal details that won't make sense to anyone. I don't see what useful information we'd provide in debug logs for the parser.

You still haven't answered my question though. What motivated this pull request? Adding it without actually needing it for yourself would just make this patch a bunch of unnecessary noise.

src/lib.rs
```diff
@@ -52,7 +52,7 @@ const MAX_OSC_RAW: usize = 1024;
 ///
 /// Generic over the value for the size of the raw Operating System Command
 /// buffer. Only used when the `std` feature is not enabled.
-#[derive(Default)]
+#[derive(Debug, Default, Clone)]
```
chrisduerr, line 55: We should implement all the standard traits that are useful to potential consumers:
https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits

I think eq/clone would probably be useful for testing. Ord/Hash shouldn't be necessary.

Debug doesn't seem relevant, I don't see how we'd provide a useful debug implementation for it that doesn't just dump a bunch of internal state that can change at any time.

chrisduerr (changes requested):

src/lib.rs
nixpulvis: Nobody should be expecting `Debug` information to be a reliable API. It's more for... well, debugging. Which I do think "leaking" internal state is perfectly reasonable to do. If we can clean up the format to be more useful, then that's great, but I don't see an issue with the derived one personally.

chrisduerr: > Which I do think "leaking" internal state is perfectly reasonable to do.

That internal state has no information that is of use to consumers though and none of it is accessible to them in any other way. It's basically just arbitrary data that has no clear meaning to library users. Littering that out to a debug implementation doesn't help anyone.

nixpulvis: I suppose I see what you mean.

It may be worth putting a little thought into what might actually be useful to users while debugging their use of the parser tho. Like how far into a sequence they are, state machine information, etc.

chrisduerr:
Closing due to inactivity.

PR #144, opened 2025-10-01
Exclude test data from being published to crates.io

weiznich:
These data are "binary" so somewhat hard to review. Also they increase the package size by quite a bit. This changes reduces the package size from 17 files, 316.5KiB (56.8KiB compressed) to 10 files, 155.0KiB (34.1KiB compressed) which results in a ~34GB/month traffic reduction for crates.io assuming the current 1.5 million downloads per month.

I'm also happy to switch this to an exclude based approach if you rather prefer that variant.


weiznich:
The build seem to fail due to https://github.com/rust-lang/rust/pull/138907. I'm happy to push a fix for that as well

Issue #146, opened 2026-02-25
missing docs for ansi feature


dsal3389:
Hi, current documentation in docs.rs is completely missing the "ansi" feature documentation 
https://docs.rs/vte/0.15.0/vte/

chrisduerr:
Happy to accept a patch, iirc getting features to show up on docs.rs can get a bit messy.

dsal3389:
opened a PR to label the ANSI feature in docs

to be honest I am not sure why the "ansi" feature is not visible in the current docs, it all looks fine

I also ran couple of [tests](https://docs.rs/dsal3389-feature-test-doc/0.1.4/dsal3389_feature_test_doc/) of my own and looks like docs.rs picked up the crate features (even without using `#[cfg_attr(docsrs, doc(cfg(feature = ...)))]`

chrisduerr:
Closing this for now under the assumption that the next release should get this fixed.

PR #147, opened 2026-02-27
docs label `ansi` feature

dsal3389:
[issue](https://github.com/alacritty/vte/issues/146)


chrisduerr (changes requested):
Have you looked into solutions for this outside of just trying to add this attribute?

Some very basic online research suggests you'll need something like his:
https://github.com/rust-random/rand/blob/master/Cargo.toml#L20-L24

I strongly recommend looking at a few projects where this is working just fine to understand how this should work.

dsal3389:
checked out `tokio` and looks like the vte Cargo.toml is correct
https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml#L168

followed this comment by Tokio maintainer:
https://users.rust-lang.org/t/how-to-document-optional-features-in-api-docs/64577/3

also mentioned in the issue's comment:
> I also ran couple of [tests](https://docs.rs/dsal3389-feature-test-doc/0.1.4/dsal3389_feature_test_doc/) of my own and looks like docs.rs picked up the crate features (even without using #[cfg_attr(docsrs, doc(cfg(feature = ...)))]

the rustdoc feature `--generate-link-to-definition` in the "rand" crate enables jumps to definition in the source code like modern editors do
image

based on the `rand` crate toml and my own testing, it is possible to remove the `rustdoc-args = ["--cfg", "docsrs"]`, and `docs.rs` should pickup the features automatically when publishing, should I maybe follow the `rand` crate instead?

or maybe it is even possible to rebuild the docs of the latest version in crates.io and the features will appear
image

chrisduerr:
Honestly I completely forgot we already added all this other stuff to the crate. Surprised it's not working already.

Let's just go with this and if it is still causing issues with the next release we can do a follow-up.

chrisduerr (approved):

chrisduerr:
Oh I just realized why I couldn't find the previous changes for me locally. They were just recently contributed *before* the last release: https://github.com/alacritty/vte/commit/ca3dd62eb4a79c9c728bdb6ec67a853cfbb801bf

So there's a good chance things would have worked out without this patch, but I don't think this will hurt.

chrisduerr:
Actually that patch introduced `#![cfg_attr(docsrs, feature(doc_auto_cfg))]`, **not** `doc_cfg`, making this patch invalid. I should have looked into this more before merging, sorry.

chrisduerr:
And to double-back *again*, the `doc_auto_cfg` was later removed in https://github.com/alacritty/vte/commit/c94b6e3060a481ae0ddfb90607323109ec2b62f4, since it is a nightly feature. Which means this patch is indeed necessary and everything is fine. :smile:

PR #152, opened 2026-05-29
Add OSC 133 prompt markers to ansi Handler

nixpulvis:
Route `OSC 133` `;A`, `;B`, `;C`, and `;D` to four new `Handler` methods: `prompt_start`, `prompt_end`, `command_start`, and `command_end`, so emulators can track shell prompt and command regions. `command_end` parses the optional exit code from the `OSC 133 ; D ; <code>` form. All four methods default to no-ops, so existing consumers are backwards compatible.

OSC 133 is emitted by shells, e.g. fish or by zsh with the `vendor_conf.d` integrations and bash with `bash-preexec`, starship, and other integrations. The four sub-commands delimit prompt and command regions and let the terminal track shell state in this way. Other terminals like kitty and ghostty also consume this protocol.

The primary motivation for this for me is https://github.com/alacritty/alacritty/pull/8961.

### DIFF
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -247,6 +247,10 @@ struct ProcessorState<T: Timeout> {
 
     /// State for synchronized terminal updates.
     sync_state: SyncState<T>,
+
+    /// Set on `command_start` or `command_executed` and cleared on
+    /// `command_finished`. Used to ignore `command_finished` when not set.
+    command_started: bool,
 }
 
 #[derive(Debug)]
@@ -699,6 +703,18 @@ pub trait Handler {
 
     /// Set hyperlink.
     fn set_hyperlink(&mut self, _: Option<Hyperlink>) {}
+
+    /// Mark the start of the prompt.
+    fn prompt_start(&mut self) {}
+
+    /// Mark the start of the command the user types, after the prompt.
+    fn command_start(&mut self) {}
+
+    /// Mark that the command was executed and its output begins.
+    fn command_executed(&mut self) {}
+
+    /// Mark that the command finished, with its exit status when known.
+    fn command_finished(&mut self, _: Option<u8>) {}
 
     /// Set mouse cursor icon.
     fn set_mouse_cursor_icon(&mut self, _: CursorIcon) {}
@@ -1520,6 +1536,33 @@ where
             // Reset text cursor color.
             b"112" => self.handler.reset_color(NamedColor::Cursor as usize),
 
+            // Prompt and command markers.
+            b"133" => {
+                if params.len() < 2 {
+                    return unhandled(params);
+                }
+                match params[1] {
+                    b"A" => self.handler.prompt_start(),
+                    b"B" => {
+                        self.state.command_started = true;
+                        self.handler.command_start();
+                    },
+                    b"C" => {
+                        self.state.command_started = true;
+                        self.handler.command_executed();
+                    },
+                    b"D" => {
+                        if self.state.command_started {
+                            self.state.command_started = false;
+                            let exit_code =
+                                params.get(2).and_then(|p| str::from_utf8(p).ok()?.parse().ok());
+                            self.handler.command_finished(exit_code);
+                        }
+                    },
+                    _ => unhandled(params),
+                }
+            },
+
             _ => unhandled(params),
         }
     }
@@ -2046,6 +2089,15 @@ mod tests {
         identity_reported: bool,
         color: Option<Rgb>,
         reset_colors: Vec<usize>,
+        prompt_events: Vec<PromptEvent>,
+    }
+
+    #[derive(Debug, PartialEq, Eq)]
+    enum PromptEvent {
+        PromptStart,
+        CommandStart,
+        CommandExecuted,
+        CommandFinished(Option<u8>),
     }
 
     impl Handler for MockHandler {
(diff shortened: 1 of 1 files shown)


src/ansi.rs
```diff
@@ -1520,6 +1532,24 @@ where
             // Reset text cursor color.
             b"112" => self.handler.reset_color(NamedColor::Cursor as usize),
 
+            // Shell integration.
```
chrisduerr: I don't like calling this shell integration, because that is very unclear and not really descriptive of what this does. I'd just refer to it as prompt markers I think.

src/ansi.rs
```diff
@@ -700,6 +700,18 @@ pub trait Handler {
...
+    fn prompt_start(&mut self) {}
+
+    /// End the shell prompt region and start a user command line region.
+    fn prompt_end(&mut self) {}
+
+    /// End the user command line region and start the command's output region.
+    fn command_start(&mut self) {}
+
+    /// End the command's output region.
+    fn command_end(&mut self, _exit_code: Option<i64>) {}
```
chrisduerr: Only values in the range of `0-255` are valid for the exit code according to iterm's docs. Why pick an i64 here?

chrisduerr: > If neither FTCS_COMMAND_START nor FTCS_COMMAND_EXECUTED was sent prior to FTCS_COMMAND_FINISHED it should be ignored.

Do you intend to let terminal emulators handle this themselves? Probably fine considering that it doesn't seem particularly bad to violate this part of the spec.

chrisduerr (changes requested):
I'm willing to merge this as long as *any* terminal emulator making use of VTE is interested in this, even if Alacritty does not receive support.

If it turns out OSC 133 is not useful to Alacritty, I'd imagine cosmic-term would still be interested in this (and a possible API in alacritty_terminal maybe).

src/ansi.rs
nixpulvis: 100% agree.

src/ansi.rs
nixpulvis: I actually wasn't sure so I was being conservative, I'll update this.

nixpulvis: Do we have a good way to track this state, I guess it would be better to follow the spec here, I was just doing the simplest thing and honestly not thinking as much as perhaps I should have about the spec here.

chrisduerr: We could store it on the parser, but honestly I'm not sure we need to account for this. It feels more like a "there's no need to handle this edgecase" warning.

If there's nothing *sensible* to do, terminals will probably have no trouble figuring that out during implementation.

I guess the biggest risk here is that a terminal would interpret this as a continuation of the last start sequence, thereby extending it. Considering the state of the ansi module should essentially be free, I think it's probably best to handle this explicitly to avoid this issue.

nixpulvis: Added a `command_started: bool` to track and implement this.

nixpulvis:
Review comments should be addressed, and I've also updated the naming of the handler methods to actually match the spec and be much less confusing.

> and a possible API in alacritty_terminal maybe

I was thinking a bit more about this, and regardless of the state of https://github.com/alacritty/alacritty/pull/8961 or our intentions of merging it, changing it significantly, or just leaving it, I think a better API to the underlying OSC 133 regions would be nice. I'm not sure if this is the right thread for it though, so I'll try to post some thoughts in https://github.com/alacritty/alacritty/issues/5850 soon.

src/ansi.rs
```diff
@@ -247,6 +247,10 @@ struct ProcessorState<T: Timeout> {
 
     /// State for synchronized terminal updates.
     sync_state: SyncState<T>,
+
+    /// Set on `command_start` or `command_executed` and cleared on
+    /// `command_finished`. Used to ignore `command_finished` when not set.
```
chrisduerr, line 252: This documentation is not particularly helpful unless you already know what it's talking about.


```suggestion
    /// Prompt marker state.
    ///
    /// Used to track whether we're inside a prompt region with marked boundaries. See [`Handler::command_finished`].
```

src/ansi.rs
```diff
@@ -1520,6 +1536,33 @@ where
...
+                    return unhandled(params);
+                }
+                match params[1] {
+                    b"A" => self.handler.prompt_start(),
+                    b"B" => {
+                        self.state.command_started = true;
+                        self.handler.command_start();
+                    },
+                    b"C" => {
+                        self.state.command_started = true;
```
chrisduerr, line 1551: This is only necessary if execution is possible without start, right? iterm2's document suggests this is not allowed, though it's not entirely clear about that.

Generally it doesn't really cover edgecases very well…

src/ansi.rs
```diff
@@ -1520,6 +1536,33 @@ where
...
+                        self.handler.command_executed();
+                    },
+                    b"D" => {
+                        if self.state.command_started {
```
chrisduerr, line 1555: I see no reason not to merge these two statements?

chrisduerr (changes requested):

src/ansi.rs
nixpulvis: We'll get an `unhandled` call when `!command_started` which I don't think we want.

chrisduerr: Ah that's fair, I forgot about that.

src/ansi.rs
nixpulvis: It says this:

> All text between FTCS_COMMAND_START and FTCS_COMMAND_EXECUTED at the time FTCS_COMMAND_EXECUTED is received excluding terminal whitespace is considered the command the user entered.

Which I don't interpret as a strict requirement, but more of an "if there is text, it's interpreted this way" kind of statement.

It then goes on to say this:

> It is expected that user-entered commands will be edited interactively, so the screen contents are captured without regard to how they came to contain their state. If the cursor's location is before (above, or if on the same line, left of) its location when FTCS_COMMAND_START was received, then the command will be treated as the empty string.

Which explicitly allows for the case of a command being empty, even if it's not directly about sending `C` without `B`.

Given that, I feel like it's reasonable to allow for this, though I really don't have a strong feeling about it right now.

I do feel like I need to do a survey of existing OSC 133 supporting shells and applications.

chrisduerr: To me the statement regarding `FTCS_COMMAND_START` suggests that you need to know about the command start to know whether it's empty or not. If you don't have this information, you won't be able to tell whether it should be empty or not.

It's fair to just *assume* it shouldn't be empty if you have no information, but it's not entirely clear (imo).

